PR #8, opened 2016-09-24
add ru_ru fruits

ranwise:


lk-geimfari:
@ranwise Please, remove vegetables and berries from list, because they use different functions for getting data.

Issue #3, opened 2016-09-13
Need support of other locales.


lk-geimfari:
We need to add:

- 🇷🇴 Roumanian
- 🇱🇹 Lithuanian
- 🇱🇻 Latvian

Feel free to send us request for another language which not in list.

**You should use**: [locale_template](https://github.com/lk-geimfari/mimesis/tree/master/mimesis/data/locale_template)

lokhi:
I can help you with french if you want

lk-geimfari:
@lokhi If something is missing, then add, please. If something in data is a wrong then fix this, please. Thank you very much!

lokhi:
Thanks I will check that ;)

auyer:
I can help with Brazilian Portuguese !

lk-geimfari:
@auyer It would be awesome! I [added](https://github.com/lk-geimfari/church/tree/master/church/data/pt-br) folder for pt-br locale. Please, check out what data we need. Thank you for your contribute!

Hellowlol:
much of the data from countries  can be found on Wikipedia. Like list of cites, names etc. would be nice to use a script to fix this

lk-geimfari:
@Hellowlol Yeah, i use `BeautifulSoup` for grabbing some data from another sites. For grabbing from Wikipedia i use Chrome [extension](https://chrome.google.com/webstore/detail/copytables/ekdpkppgmlalfkphpibadldikjimijon) that helped copping a column from a table.

jackmcmorrow:
Hey @auyer ! I want to help with ptbr too! Hit me up and we can work on it together!

jackmcmorrow:
Here is a list of brazilian cities with more than 100.000 inhabitans. That's 300+ names in a table.

jackmcmorrow:
Oops, forgot the link https://pt.m.wikipedia.org/wiki/Lista_de_municípios_do_Brasil_acima_de_cem_mil_habitantes

auyer:
@jackmcmorrow Sure ! Thanks ! I'll push my changes to my fork. You can make PRs there ! https://github.com/auyer/church

yebrahim:
Are you looking for RTL languages too?

lk-geimfari:
@yebrahim Yes, but I do not know how to work with these data. So, if you know how to add data for RTL locale correctly then i support your decision.

theomessin:
Want to help with Greek!

lk-geimfari:
[...]

PR #21, opened 2016-10-16
Update Spanish locale

Costava:
Add to and update the Spanish locale.

Issue #3


lk-geimfari:
@Costava Thank you very much!

PR #29, opened 2016-10-20
Added sphinx configuration

Battleroid:
I added the sphinx configuration, tried my best to make the docstrings throughout more consistent with what sphinx expects (so it doesn't throw a fit), as well as included the readme and guidebook material.

Please let me know if there's anything I've goofed, needs changing, etc.


lk-geimfari:
@Battleroid Great job! Thanks!

Issue #38, opened 2016-10-24
Add abbreviation to State field


memilanuk:
Add option for abbreviation of state name i.e. something like:

`address.state(abbrev=True)`

to return 'WA' for state, vs. 'Washington'.  

Probably want to do something similar for states/provinces in other countries?

lk-geimfari:
I think that it's possible. We will add this functionality soon. Thanks!

lk-geimfari:
@memilanuk [It's](https://github.com/lk-geimfari/church/commit/854c6be106897ff9ba0aaa7dacfa8e8749a5eeaf) done!

PR #40, opened 2016-10-25
Abbrev

mlterpstra92:
I have added a flag to state to select the abbreviation of a state rather than the full name as per Issue #38. It is tested and the flag defaults to False.


lk-geimfari:
@mlterpstra92 I already did it. Thank you!

PR #43, opened 2016-10-25
Added locale specific test cases and fix minor key error with chemical_element

Battleroid:
I managed to rewrite a the tests a little bit and add some test cases so all locales or individual ones can be tested.

Also found while testing languages other than english that some of the science and data json data files had the `chemical_element` key as `chemical_elements` occasionally so I changed that, all the tests passed once changed.

Let me know if I goofed anything. I'm sure there's probably a better way to write the unit tests, but they do work, so maybe for now this might be okay.


lk-geimfari:
@Battleroid The `data.json` is unused file.

PR #46, opened 2016-10-27
Fixes #45 (Wrong display of a french address)

### LINKED ISSUE #45: Wrong display of a french address
```
>>> from church import Address
>>> addr = Address('fr')
>>> addr.address()
'371 Bezout Rue du'
```

The output should be:
'371 Rue du Bezout'

mirelsol:


lk-geimfari:
It's already fixed.

Issue #51, opened 2016-11-08
Solve the issues that arose after the renaming of the library.


lk-geimfari:
The library has been renamed (I hope that this name is not offensive) and now we need to check all code lines of our library.

TODO:
- Check documentation.
- Check docstrings.

Battleroid:
Submitted #52.

lk-geimfari:
@Battleroid Thank you!

Uncleleech:
Submitted #69, unsure if you guys wanted it or not so feel free to reject.

PR #52, opened 2016-11-12
Fix up name in index.rst

Battleroid:
From what I can tell this is all that needs to be changed due to the rename (#51).


lk-geimfari:
@Battleroid Thanks!

PR #54, opened 2016-11-24
Update utils.py

offermann:
On Windows, the file encoding 'utf8' has to be specified explicitly.


lk-geimfari:
@offermann I never tested this library on Windows. Thanks you very much! 

P.S If you find any more problems on your platform, then let me know, please.

PR #67, opened 2016-12-21
Added CPF and CNPJ to the Code class

martini97:
The CPF and CNPJ are codes use for identifying brazilian persons and
companies, since it requires some calculation it cannot be generated
with the custom code

Signed-off-by: Alessandro Martini <alessandrofmartini@gmail.com>

### DIFF
--- a/elizabeth/core/elizabeth.py
+++ b/elizabeth/core/elizabeth.py
@@ -551,6 +551,74 @@ class Code(object):
         """
         return self.custom_code(mask=mask)
 
+    @staticmethod
+    def cpf(with_mask=True):
+        """
+        Get a random CPF (brazilian social security code)
+        :param with_mask: use CPF mask (###.###.###-##) in the return
+        :returns: Random CPF
+        """
+        def get_verifying_digit_cpf(cpf, peso):
+            """
+            Calculates the verifying digit for the cpf
+            :param cpf: list of integers with the cpf
+            :param peso: integer with the weigth for the modulo 11 calcule
+            :returns: the verifying digit for the cpf
+            """
+            soma = 0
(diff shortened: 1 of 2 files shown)


elizabeth/core/elizabeth.py
```diff
@@ -551,6 +551,74 @@ def pin(self, mask='####'):
         """
         return self.custom_code(mask=mask)
 
+    @staticmethod
+    def cpf(with_mask=True):
+        """
+        Get a random CPF (brazilian social security code)
+        :param with_mask: use CPF mask (###.###.###-##) in the return
+        :returns: Random CPF
+        """
```
lk-geimfari: It would be like that:

```
 def cpf(with_mask=True):
         """
         Get a random CPF (brazilian social security code)
         :param with_mask: use CPF mask (###.###.###-##) in the return
         :returns: Random CPF
         :Example:
              Example here.
         """
```

elizabeth/core/elizabeth.py
```diff
@@ -551,6 +551,74 @@ def pin(self, mask='####'):
...
+        def get_verifying_digit_cpf(cpf, peso):
+            """
+            Calculates the verifying digit for the cpf
+            :param cpf: list of integers with the cpf
+            :param peso: integer with the weigth for the modulo 11 calcule
+            :returns: the verifying digit for the cpf
+            """
```
lk-geimfari: Here too, please.

lk-geimfari:
@martini97 Great work! Thanks!

lk-geimfari:
@martini97 Please, refactoring. All variable names should be in English and only in English.

PR #71, opened 2016-12-22
Path tests are now run without failures on Windows.

mipaaa:


tests/test_data/test_path.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-import sys
+import sys, os
```
lk-geimfari: New import on new line, please. We love `PEP8`.

```python
import os
import sys
```

mipaaa: Done

lk-geimfari: @mipaaa Thank you!

PR #72, opened 2016-12-24
Fix flask-church link in docs

cl0ne:


lk-geimfari:
@cl0ne Thank you!

Uncleleech:
I saw this, I think it's the only other mention of church in the program from a search

PR #76, opened 2016-12-27
Added 'add_provider' to generic provider

Battleroid:
Issue #74

`add_provider()` on Generic adds an instance of the provider class, so a new provider class such as `my_generic.add_provider(SampleProvider)` will let you call `my_generic.SampleProvider.method()` afterwards.

I used the class name to set the attribute, but it could definitely be changed to look for a key variable in the new provider to set the attribute within the generic provider.

Raises TypeError if a class type is not passed.

### DIFF
--- a/elizabeth/core/elizabeth.py
+++ b/elizabeth/core/elizabeth.py
@@ -8,6 +8,7 @@
 import os
 import sys
 import array
+import inspect
 from datetime import (
     date,
     timedelta,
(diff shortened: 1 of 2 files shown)


lk-geimfari:
@Battleroid Great job! You are making a huge contribution to Elizabeth. I hope that you will continue in the same spirit. Thank you!

Battleroid:
No problem. Glad I could help.

PR #77, opened 2016-12-27
Added check for meta class to set provider name

Battleroid:
Added the ability to set the provider name via a meta subclass. If the subclass isn't available the name will be set using the standard way (`cls.__name__.lower()`).

Modified the test to use an alternate name to test the functionality.


lk-geimfari:
@Battleroid I added it just now. :smile:

Battleroid:
Woo!

PR #83, opened 2017-01-02
add personality generation

aburgd:
Authors could use `elizabeth` for character information, so I figured to add personality should be included. The personality objects in some of the non-English locales need translation, but the test passes in `test_personal.py`, and pycodestyle found no errors in the added function.


elizabeth/data/da/personal.json
```diff
@@ -2022,5 +2022,36 @@
...
+		  "ESTP",
+		  "ESFP",
+		  "ENFP",
+		  "ENTP",
+		  "ESTJ",
+		  "ESFJ",
+		  "ENFJ",
+		  "ENTJ"
+	  ],
+	  "rheti": [
```
lk-geimfari: It's duplicated in all 17 locales. Digits can be generated in method `personality()`

For example:

```python
if category == 'rheti':
    return randint(1, 10)
```
Please, fix it.

lk-geimfari: Or you can move it to specific data providers. Class `USA` in elizabeth.builtins.

lk-geimfari: Look [here](https://github.com/lk-geimfari/elizabeth#builtins-specific-data-providers)

lk-geimfari:
@aburgd It's done!

Issue #79, opened 2016-12-28
Add comparison with Faker 


lk-geimfari:
Is an commonly asked question. We need to compare Fake and Elizabeth.

Uncleleech:
as I said in #75, we can def use a speed comparison as in #12.

We dont need to compare what is better so much as what is different I mean that, if faker has some behaviour that elizabeth doesnt show how it differs (the same vice versa)

Also I can see mentioning the structure of the locale files might be good (applying the KISS paradigm :) as I think theirs is a bit of a mess)

lk-geimfari:
@Uncleleech I completely agree with you. I will add a small comparison of data generation to the README.  Comparison for persons that always ask me about Faker.

Battleroid:
Sorry, I responded to the wrong issue, meant to post it here:

Yes, I think it's useful. It's be nice to have an in depth comparison vs faker, both with usage examples and speed comparisons.

fcurella:
I'd like to expand the comparison with Faker, but I also think it should be moved higher in the README. Not necessarily at the top, but probably before installation and usage.

My understanding is that main advantage of this project is speed. Was there anything else that's not available on Faker or that Elizabeth does better / differently?

lk-geimfari:
Speed. More specific data. As i say i did not carry out in-depth comparison, because I used the only way Faker:
```
from faker import Faker
faker = Faker()

faker.name()
faker.email()
```
Nothing more. But if you can compare it would be cool.

PR #84, opened 2017-01-02
#79 Expand comparison section

fcurella:
ref #79


lk-geimfari:
@fcurella Great! Thank you!

PR #85, opened 2017-01-05
fixed issue with wrong method name in documentation

wikkiewikkie:
Person.occupation() method was referred to as Person.profession() in examples.


lk-geimfari:
@wikkiewikkie Thank you!

PR #86, opened 2017-01-05
Add Python compatibility note in Faker comparison

fcurella:


lk-geimfari:
@fcurella We already have badge about supported Python version  in top of README.

lk-geimfari:
@fcurella Anyway, thank you.

PR #87, opened 2017-01-05
refactor password method

wikkiewikkie:


lk-geimfari:
@wikkiewikkie Thanks!

PR #88, opened 2017-01-05
Improvements to Datetime

wikkiewikkie:
Removed the birthday method which seemed redundant of date.
Updated date and time methods to use locale-specific formats and refactored them to resolve a few issues:

- date() would only return dates with a day <= 28
- sep and with_time parameters consolidated into a format string.

Updated tests and docs to reflect all of the above.


lk-geimfari:
@wikkiewikkie Very good job. Thank you.

Issue #60, opened 2016-12-17
Update documentation.


lk-geimfari:
Documentation is outdated. We should fix it immediately.

PR #92, opened 2017-01-06
reorganize and update README

wikkiewikkie:
#60


lk-geimfari:
@wikkiewikkie It's look much better. Thank you.

PR #94, opened 2017-01-06
Add returning a random SSN to USA provider

drpoggi:
I removed the geographical area codes because according to [this](https://www.ssa.gov/employer/randomization.html) they are no longer relevant. Let me know if you'd like anything changed. Thanks!


lk-geimfari:
@drpoggi Thanks!

PR #95, opened 2017-01-07
add ko locale data

redus:
here's data for ko locale, i'll commit actual usage pretty soon.


lk-geimfari:
@redus Thank you. Can i merge it? That's all?

lk-geimfari:
@redus Tell me, please where are you from? Which flag we can use for locale `ko`  in README? South Korea?

redus:
from South Korea, and yep south Korean flag for ko. just adding data so shouldn't break anything.

lk-geimfari:
@redus Now i  will run tests. I'll let you know about the result.

redus:
<streetname><street> <number>
is the most common case

redus:
oh it's not showing umm 
[street-name][street] [number]

lk-geimfari:
@redus Good. Thanks!

redus:
yup, I'll improve on address data later, too.

PR #99, opened 2017-01-07
update for ko postal codes

redus:
we've gone through postal code pattern change in near 2015, and i've updated the pattern and tests for it. all tests passing on my end. Thanks for all,


lk-geimfari:
@redus I'm sorry i didn't know about this. Thank you for PR.

redus:
failing on chemical name test. Korean has elements where the length is less than two, so yeah could be exceptioned with farsi as well

redus:
yup no problem

lk-geimfari:
@redus I fix issue with length of chemical element as son as possible.

PR #105, opened 2017-01-10
Structured Data Provider

wikkiewikkie:
Allows for generation of structured data.  Currently includes CSS, Delimited, and HTML methods.


drpoggi:
What's the use case for this? I'm not against it at all, just curious.

wikkiewikkie:
@drpoggi HTML and delimited I think would be useful if you are developing something that allows users to input HTML (CMS, for example) or to import delimited date (database front-end/spreadsheet app).
Faker also offers a HTML lorem provider that generates a complete HTML document and this is an initial step towards implementing something more like that.   Might be useful if you are developing say, a search engine that indexes HTML documents.

drpoggi:
Ok makes sense, thought it might be something like that, thanks for the explanation!

lk-geimfari:
@wikkiewikkie I'm not sure that html tags are useful, but i will merge it, because i think thtat `delimited()` is very useful. Also it would be great, if you add to Structured `jsonify()` that will write data to json file.

lk-geimfari:
@wikkiewikkie We need refactoring of Structured. PEP8.

PR #109, opened 2017-01-18
Store generated object field value to generate depended fields - example for Personality.age

yn-coder:
Store generated object field value to generate depended fields - example for Personality.age

https://github.com/lk-geimfari/elizabeth/issues/107

With test coverage

### DIFF
--- a/elizabeth/core/elizabeth.py
+++ b/elizabeth/core/elizabeth.py
@@ -803,9 +803,10 @@ class Personal(object):
         """
         self.locale = locale
         self.data = pull('personal.json', self.locale)
+        self.stored_age = 0
 
-    @staticmethod
-    def age(minimum=16, maximum=66):
+    #@staticmethod
+    def age(self, minimum=16, maximum=66, use_stored = False ):
         """
         Get a random integer value.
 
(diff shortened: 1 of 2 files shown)


elizabeth/core/elizabeth.py
```diff
@@ -815,8 +816,42 @@ def age(minimum=16, maximum=66):
         :Example:
             23.
         """
-        return randint(minimum, maximum)
+        a = randint(minimum, maximum)
+        self._store['age'] = a
+        return a
 
+    def child_count( self, max_childs = 5 ):
```
lk-geimfari: Should be: `def child_count(self, max_childs=5):`.

elizabeth/core/elizabeth.py
```diff
@@ -815,8 +816,42 @@ def age(minimum=16, maximum=66):
...
+        if a == 0:
+            a = self.age()
+        
+        if a < 18:
+            cc = 0 # underage
+        else:
+            cc = randint( 0, max_childs )
+        return cc
+    
+    def work_experience( self, working_start_age = 22 ):
```
lk-geimfari: Here too.

elizabeth/core/elizabeth.py
lk-geimfari: and here

lk-geimfari:
Read about PEP8, please. Merged, but in future you should know about [PEP8](https://www.python.org/dev/peps/pep-0008/).

yn-coder:
You mean `too many blancks` ?

lk-geimfari:
@yn-coder Yes. I recommend you really good [guide](http://docs.python-guide.org/en/latest/).

PR #114, opened 2017-01-22
add jp locale

jasonwaiting-dev:
Here's the locale data for jp. 🇯🇵
(I am not Japanese but lived in Japan for a couple of years.)


lk-geimfari:
@jasonwaiting-dev Thank you! And tell me please following things:
1. What is postal code format in Japan.
2. What is Japan international phone number code.
3. What is Japan ISBN group code.
4. What is address format in Japan.

For example in Russia we have following format `{siffux} {street_name} {street_number}`

lk-geimfari:
Also run tests, please. Just add class to `tests/test_locales.py`

```
class JapaneseTest(LocaleBase, TestCase):
    LANG = 'jp'
```

jasonwaiting-dev:
@lk-geimfari Thanks for your feedback! I

> What is postal code format in Japan.

###-####, e.g. 100-0001

> What is Japan international phone number code and format of phone number.

+81 ### #### ####

> What is Japan ISBN group code.

4

> What is address format in Japan.

Japanese addressing system is very different from other countries. We basically use district + block + building only.
For example,
> 〒100-8994
Tōkyō-to Chūō-ku Yaesu 1-5-3
Tōkyō Chūō Yūbin-kyoku

|Postal code|
|Prefecture| |Ward| |Town| |District|-|Block|-|Building|
|Building Name|
[reference](http://www.sljfaq.org/afaq/addresses.html)

Most streets in Japan don't even have names. The street names I used in address.py are all the main streets in the cities. 
The current address format I am using is `{street_name}{street_number} {suffux}` so that all tests can be passed. However, I think we have to address this issue! I noticed that there are some customizations for different locales, e.g.
```python
if self.locale == 'ru':
    nations = self.data['nationality'][gender]
    return choice(nations)
else:
    return choice(self.data['nationality'])
```
Should we do the same for Japan addressing system?

lk-geimfari:
@jasonwaiting-dev If this is one correct way then yes, we should.

PR #115, opened 2017-01-24
add town for JP addressing system

jasonwaiting-dev:
regarding the Japanese addressing system, we didn't handle it well in the current JP locale.
lk-geimfari/elizabeth#114

I added an exceptional case for jp address. 
using format: {Town} {District}-{Block}-{Building}
- Town: from a list of towns in jp/address.py
- District: random number (max = 100)
- Block: random number (max = 100)
- Building: random number (max = 100)

All tests are passed without error. :D


lk-geimfari:
@jasonwaiting-dev Great job! どうもありがとうございます :smile:

lk-geimfari:
@jasonwaiting-dev Can i merge it?

PR #116, opened 2017-01-24
removed duplicates from data

aybb:
I found some duplicates in russian fixtures ( personal / occupation ), but decided to verify all fixtures at all locales.


lk-geimfari:
@aybb Sorry, here to many changes. I can't check where removed duplicates and where deleted non duplicated data. Can you resend PR, but with separated commits by locale?

Example of commits:
```
Removed duplicates from ru
Removed duplicates from en
Removed duplicates from de
# ...
```
Thanks.

drpoggi:
@aybb it looks like your editor re-formatted most (all?) of the locale files, which created a very large PR for what doesn't look to be too many actual changes.

aybb:
Ye, but script i used looks pretty clean and don't suppose to delete non-duplicate fixtures. Also there's broken indentation at some locales and removing duplicates from this lines might be painful.

lk-geimfari:
@aybb Using script for this problem is insecure. Yes, we can use script for finding duplicates, but we should delete duplicates manually. It's a safe way.

Issue #81, opened 2016-12-29
Add builtins specific data providers.


lk-geimfari:
Every language has specific data that suit only for ones. For example `SSN` for `en` (USA) or `CPF` for `pt-br`. `CPF` can be useful only for brazilians. 

If user want to use this providers then he must be imported explicitly.

Here's how it will look:
```python
>>> from elizabeth import Generic
>>> from elizabeth.builtins import Brazil

>>> generic = Generic('pt-br')

>>> class BrazilProvider(Brazil):
        class Meta:
            name = "brazil_provider"
```
```python
>>> generic.add_provider(BrazilProvider)
>>> generic.brazil_provider.cpf()
'001.137.297-40'
```

lk-geimfari:
@Battleroid What you think about it? It's sounds useful?

Uncleleech:
Yes, this is useful.
Perhaps we can also make a general one for custom pieces of data that can be useful as well altough im having trouble picturing how you'd make that

Battleroid:
Yes, I think it's useful. It's be nice to have an in depth comparison vs faker, both with usage examples and speed information.

lk-geimfari:
@Battleroid @Uncleleech Also i [added](https://github.com/lk-geimfari/elizabeth#comparison) really small comparison to README with screenshot.

lk-geimfari:
@Battleroid If you wish, you can carry out more in-depth comparison and add the results to the README. My comparison was too humble.

Battleroid:
Sounds good.

PR #120, opened 2017-01-28
Added converter between half-width and full-width for jp

redus:
#81 and #3 
Hope this is sufficient for the jp locale usability


lk-geimfari:
@redus It's looks beautiful. Thank you!

PR #128, opened 2017-02-13
Fix regexp in tests/test_internet.py

### LINKED ISSUE #126: test_internet: dubious regexp
`tests/test_data/test_internet.py` includes the following regexp fragment:
```
[$-_@.&+]
```
But `$-_` is a character range that includes digits, uppercase letter and a bunch of punctuation characters.
You probably wanted this instead:
```
[$_@.&+-]
```

The dubious regexp was found using [pydiatra](https://github.com/jwilk/pydiatra).

jwilk:
Fixes #126


lk-geimfari:
@jwilk Please, add your name to list of contributors. Thanks.

PR #133, opened 2017-02-15
improve test coverage and test assertions

wikkiewikkie:


lk-geimfari:
@wikkiewikkie Please, add action that will remove downloaded image `elizabeth_1.png` (tear_down). Thanks.

Issue #132, opened 2017-02-15
Compatibility issue with utils.download_image()


wikkiewikkie:
I am trying to write a test to cover situations where unverified_ctx is true and the build is failing on 3.3 and 3.4 with the following error:

`E           AttributeError: 'module' object has no attribute '_create_unverified_context'`

I did some research and discovered that ssl._create_unverified_context was renamed from ssl._create_stdlib_context in 3.4, but does not exist in 3.3.

Changing _create_unverified_context to _create_stdlib_context will allow the build to pass on 3.4, but it will still fail on 3.3.

See [PEP 476](https://www.python.org/dev/peps/pep-0476/) for info on ssl._create_unverified_context and ssl._create_stdlib_context.

lk-geimfari:
@wikkiewikkie Maybe we can check version and run it only in `Python 3.4^`? Any other ideas?

wikkiewikkie:
@lk-geimfari This was fixed in #136.

PR #136, opened 2017-02-16
clean up downloaded image after test

wikkiewikkie:
@lk-geimfari as requested in #133


wikkiewikkie:
@lk-geimfari this also fixes #132

PR #152, opened 2017-03-15
add zh data

jlwt90:
Hello, I finally added data for simplified chinese (china)!
Could you please check if there is any problem?

Thanks!

PR #153, opened 2017-03-18
remove trailing space in zh text section

jlwt90:
fix for #152 zh text

> Also, please. Help to understand why sometime tests fails for zh:

Sorry for that! I didn't notice that there is a trailing space. Since we `strip()` the value before assertion, the test fails when that sentence is chosen by `generic.text.sentence()`!


lk-geimfari:
@jlwt90 Thanks!

PR #155, opened 2017-03-21
Added a option to personal provider to generate a random gender

jrast:
This PR is about issue https://github.com/lk-geimfari/elizabeth/issues/154 and implements a option to generate a gender randomly if `None` is passed to any `gender` parameter on personal data.

Note: In my opinion, `None` should be de default and not `female`. But this would be a change which is not backward compatible and might break some stuff.

### DIFF
--- a/elizabeth/utils.py
+++ b/elizabeth/utils.py
@@ -5,8 +5,9 @@ import functools
 import json
 import os.path as path
 import urllib.request as request
+from random import choice
 
-from elizabeth.exceptions import UnsupportedLocale
+from elizabeth.exceptions import UnsupportedLocale, WrongArgument
 from elizabeth.settings import SUPPORTED_LOCALES
 
 __all__ = ['pull', 'download_image' 'locale_information']
@@ -146,3 +147,21 @@ def update_dict(initial, other):
         else:
             initial[key] = other[key]
     return initial
+
+def check_gender(gender):
+    """
+    Check a given gender if it's valid
+    
+    Valid genders are 'male', 'female' and None. If None is given, a gender
+    is randomly choosen.
+
+    :param gender: str or None
+    :return: One of 'male' or 'female'
+    :rtype: str
+    """
+    if gender is None:
+        return choice(['male', 'female'])
+    gender = gender.lower()
+    if gender not in ['male', 'female', None]:
+        raise WrongArgument('gender must be "female", "male" or None but was "%s"' % str(gender))
+    return gender
--- a/elizabeth/utils.py
+++ b/elizabeth/utils.py
@@ -162,6 +162,6 @@ def check_gender(gender):
     if gender is None:
         return choice(['male', 'female'])
     gender = gender.lower()
-    if gender not in ['male', 'female', None]:
-        raise WrongArgument('gender must be "female", "male" or None but was "%s"' % str(gender))
+    if gender not in ['male', 'female']:
+        raise WrongArgument('gender must be "female", "male" or None but was "%s"' % gender)
     return gender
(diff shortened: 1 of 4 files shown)


jrast:
PS:  Codacy is strange... After the initial PR it only said whitespace on empty lines is not good, now it complains about insecure usage of random/choice and useage of assert's in tests...

elizabeth/utils.py
```diff
@@ -146,3 +147,21 @@ def update_dict(initial, other):
...
+    is randomly choosen.
+
+    :param gender: str or None
+    :return: One of 'male' or 'female'
+    :rtype: str
+    """
+    if gender is None:
+        return choice(['male', 'female'])
+    gender = gender.lower()
+    if gender not in ['male', 'female', None]:
```
wikkiewikkie: None values are caught by the first if condition, so gender will never be equal to None here

elizabeth/utils.py
```diff
@@ -146,3 +147,21 @@ def update_dict(initial, other):
...
+        raise WrongArgument('gender must be "female", "male" or None but was "%s"' % str(gender))
```
wikkiewikkie: str() call is not needed

tests/test_utils.py
```diff
@@ -77,3 +78,16 @@ def test_update_dict():
 
     result = update_dict(first, third)
     assert "spaniel" not in result['animals']['dogs']
+
+
+def test_check_gender():
```
wikkiewikkie: feels like you are just testing python's random module here, not your code
also need a test to confirm WrongArgument is raised when appropriate

wikkiewikkie:
@jrast a few suggestions

elizabeth/utils.py
jrast: Yep, removed in update

elizabeth/utils.py
jrast: Removed in update

tests/test_utils.py
jrast: Yep, the test was maybe sowhat bad formulated. What do you think about the new version? Anyhow, realy test if both values `male` and `female` are actually returned, one should probably mock the `random.choice` method.

The thest for the `WrongArgument` is added

jrast:
What do you think about the suggesition of `None` as default?

lk-geimfari:
@jrast It's raise problems for some locales. For example with Russian, Icelandic, Ukrainian, because in this locales surnames separated by gender.

If we use `None` as a default value library can generate female surname for male name, but we can change it for `full_name(gender=None)`

```python
>>> _ for _ in range(0, 5):
    full_name(gender=None)
'Kenny Cain'
'Milan Mcpherson'
'Rheba Calderon'
'Linsey Mejia'
```

wikkiewikkie:
@lk-geimfari  I don't think there needs to be any relationship between `surname()` and `first_name()` and each should return a random name from a random gender unless the user asks otherwise.

To make this clearer, consider refactoring gender so it returns an [Enum](https://docs.python.org/3/library/enum.html) so it could be used like this:

```
>>> my_gender = personal.gender()
>>> type(my_gender)
<enum 'Gender'>
>>> print(my_gender)
female
>>> print(my_gender.iso5218())
2
>>> print(my_gender.symbol())
♀
>>> first_name = personal.first_name(gender=my_gender)
>>> print(first_name)
Elizabeth
```

```
>>> my_gender = elizabeth.Gender.MALE
>>> first_name = personal.first_name(gender=my_gender)
>>> print(first_name)
Christopher
```

What do you think?

lk-geimfari:
@wikkiewikkie Sounds good. Can you create PR, please?

PR #158, opened 2017-03-24
Flag to return date as object instead of a string

jrast:
It would be nice to have the option to get date as objects instead of strings. 

What do you think? If this would be fine I will add tests and documentation.


lk-geimfari:
@jrast I think that it's can be useful. Add docs and tests, please.

Issue #159, opened 2017-04-05
Publish to PyPi on merge to master


lk-geimfari:
Publishing of new release should be automated. We can do it using Travis CI.

Great example is a [`travis.yml`](https://github.com/sobolevn/django-split-settings/blob/master/.travis.yml) file of this [`project`](https://github.com/sobolevn/django-split-settings).

PR #166, opened 2017-06-02
publish to pypi

axce1:
#159 publish to pypi on create new tag


lk-geimfari:
@axce1 Great job. Thank you very much! I'll merge PR later.

PR #167, opened 2017-06-02
Adds base class for all provides

sobolevn:
Changes:
1. Removes `__str__` method from every class
2. It is now handled by new `BaseProvider` class
3. Adds `.gitignore` as industry-standard

I guess, it would also be a good idea to separate providers by different modules. Since it is almost 2k LOC. Really hard to read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,160 @@
+#### joe made this: http://goel.io/joe
+#### macos ####
+*.DS_Store
+.AppleDouble
+.LSOverride
+
+# Icon must end with two \r
+Icon
+
+
+# Thumbnails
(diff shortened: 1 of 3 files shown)


lk-geimfari:
@sobolevn Awesome! Totally agree with you. We need to separate providers by different files. The general question is how? Different file for each provider?

sobolevn:
I think a separate file will do.

Also, please take a note, that I made a breaking (lol) change. `Food` had `lang`, now it has `locale` as any other providers.

lk-geimfari:
@sobolevn These are my old mistakes. We need general refactoring.

PR #173, opened 2017-06-07
Remove bogus "axis." as Polish address component

jwilk:
I don't what what *axis.* is supposed to stand for, but we certainly don't use it in Polish addresses.


lk-geimfari:
@jwilk Thank you!

Issue #174, opened 2017-06-10
Still failed some tests


lk-geimfari:
Basically I fixed the problem with `ImportError`, but code need refactoring.

There are some problems with providers, namely: `Structured()`, `Text()`  and  `Generic()`.

Issue type:
```
TypeError: __init__() takes 1 positional argument but 2 were given
```

Full report [here](https://travis-ci.org/lk-geimfari/elizabeth/jobs/241551298#L4021)

axce1:
i'll try fix it next week

lk-geimfari:
@axce1 Thank you! Keep in mind that this error in [feature-new-strucutre](https://github.com/lk-geimfari/elizabeth/tree/feature-new-strucutre).

lk-geimfari:
I was try to fix this issue, but I could not. All test still failed. Also imports looks very ugly, after refactoring. but otherwise nothing worked.

lk-geimfari:
Above I mean branch feature-new-structure.

lk-geimfari:
@Valerievich If you still interested, you can try fix this.

Valerievich:
Sorry, but full report does not exist.

PR #176, opened 2017-06-13
Added prescription drug and currency data providers.

mawpar:


lk-geimfari:
@mawpar `currency()` is already available in providers `Business()`.

lk-geimfari:
Also i suggest create new class-provider `Pharmaceutic()` (or you can use another suitable name which you want), which will provide data related with medicine and pharmaceuticals.

lk-geimfari:
If you want to contribute also take a look at #174  issue (branch: [feature-new-strucutre](https://github.com/lk-geimfari/elizabeth/tree/feature-new-strucutre) )

lk-geimfari:
@mawpar Tell me, please. What is actually mean `currency`? Do you mean financial currency?

PR #183, opened 2017-07-01
Feature - PBKDF2 Password Support

JuxhinDB:
Adding support for PBKDF2 password algorithm since certain web apps will enforce that at an ORM level. 

##### Changes

- Added new password algorithm, PBKDF2 with 10K iterations and a random salt
- Updated relevant unittest to also test for new algorithm

All unittests passed, no additional documentation was added as it did not seem needed in this case. 

----

### DIFF
--- a/elizabeth/core/providers.py
+++ b/elizabeth/core/providers.py
@@ -993,7 +993,7 @@ class Personal(BaseProvider):
         password = ''.join([choice(
             ascii_letters + digits + punctuation) for _ in range(int(length))])
 
-        if algorithm is not None:
+        if algorithm:
             algorithm = algorithm.lower()
             password = password.encode()
             if algorithm == 'sha1':
--- a/elizabeth/core/providers.py
+++ b/elizabeth/core/providers.py
@@ -9,7 +9,8 @@ import os
 import re
 import sys
 from calendar import monthrange
-from hashlib import sha1, sha256, sha512, md5
+from binascii import hexlify
+from hashlib import sha1, sha256, sha512, md5, pbkdf2_hmac
 from random import (
     choice,
     sample,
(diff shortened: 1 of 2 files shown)


JuxhinDB:
Regarding this PR (being my first real one), I wanted to add that I intended to add `BCrypt` support as that was my use-case, however as some of you may know this is not including the the standard library and from what I can tell, the project aims to avoid third-party libraries entirely.

lk-geimfari:
@JuxhinDB Thank you for this PR. We try to avoid dependencies from third-party libraries.

JuxhinDB:
@lk-geimfari That's fine. This particular PR makes use of the std lib only.

lk-geimfari:
@JuxhinDB Please, look at Travis CI. It seems like some tests failed.

JuxhinDB:
@lk-geimfari Seems that the project supports 3.3 onward. Unfortunately Python 3.3 does not have `PBKDF2`:

    Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:18:55) [MSC v.1900 64 bit (AMD64)] on win32
    >>> import hashlib
    >>> hashlib.algorithms_available
    {'whirlpool', 'sha512', 'SHA1', 'SHA224', 'DSA-SHA', 'sha224', 'MD5', 'SHA512', 'md5', 'SHA384', 
    'dsaEncryption', 'sha', 'SHA', 'md4', 'sha256', 'sha384', 'RIPEMD160', 'DSA', 'sha1', 'ecdsa-with-SHA1', 
    'dsaWithSHA', 'MD4', 'SHA256', 'ripemd160'}

In that case, I suppose we can close this PR. :-)

lk-geimfari:
@JuxhinDB We can not disrupt backward compatibility due to one feature, sorry.

PR #191, opened 2017-07-14
TODO: Resolve importing errors and use importing format as above.

Valerievich:
In this PL: 1) all import errors were resolved 2) import format in generic.py was changed.
All tests passed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ install:
   - pip install flake8-commas
   - pip install flake8-quotes
   - pip install pytest-flake8
-  - pip install pytest-isort
 
 script: py.test --cov=mimesis/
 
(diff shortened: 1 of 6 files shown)


lk-geimfari:
@Valerievich Are you tried this format?

```python
from mimesis.providers import Games
```
Erorrs raised only when we use following format. General questions is "Why?" and "What we need to do to fix this." I think you agree that is will much better use similar format for importing all providers from `providers`.

How it should looks like:

```python
# All providers from `providers`
from mimesis.providers import (Address, BaseProvider, Business, ClothingSizes, Code, Datetime, 
                               Development, File, Food, Games, Hardware, Personal, ...)
```

Valerievich:
Yes, I tried format 
```python
from mimesis.providers import Games
```
It leads to import errors in generic.py

Valerievich:
Ok, then. I will try to use proposed format one more time and answer the question "What we need to do to fix this."

lk-geimfari:
@Valerievich Thank you!

PR #197, opened 2017-07-20
Add ukrainian text

tytarenko:


lk-geimfari:
@tytarenko Much better. Thank you very much!

PR #198, opened 2017-07-24
Completed TODO: Resolve importing errors and use importing format as above.

Valerievich:
All the tests were passed. Local build works well.

### DIFF
--- a/mimesis/providers/generic.py
+++ b/mimesis/providers/generic.py
@@ -1,20 +1,10 @@
 import inspect
 
 from mimesis.providers import (Address, BaseProvider, Business, ClothingSizes,
-                               Code, Datetime, Development, File, Food)
-
-# TODO: Resolve importing errors and use importing format as above.
-from .games import Games
-from .hardware import Hardware
-from .internet import Internet
-from .numbers import Numbers
-from .path import Path
-from .personal import Personal
-from .science import Science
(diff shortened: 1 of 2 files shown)


mimesis/providers/generic.py
```diff
@@ -1,20 +1,10 @@
...
-from .internet import Internet
-from .numbers import Numbers
-from .path import Path
-from .personal import Personal
-from .science import Science
-from .text import Text
-from .transport import Transport
-from .units import UnitSystem
-from .cryptographic import Cryptographic
+                               Code, Datetime, Development, File, Food, Games,
```
sobolevn: Could you please follow "one-line-one-import" standard?

Valerievich: Yes, I could. But current version of generic.py already breaks the rule, doesn't it? Should I refactor all lines with imports? I suppose @lk-geimfari wanted to use more compact view in this case.

lk-geimfari: @sobolevn It seemed to me that this way is more compact, but if you think that this is not the better way of importing, then we can request refactoring.

P.S Yes, this way was really requested by me. Thanks @Valerievich

sobolevn: I'm totally fine with both versions

PR #200, opened 2017-07-31
Resolved TODO: platform should be a parameter

Valerievich:


mimesis/providers/path.py
```diff
@@ -9,10 +8,10 @@
 class Path(BaseProvider):
     """Class that provides methods and property for generate paths."""
 
-    def __init__(self, *args, **kwargs):
-        # TODO: platform should be a parameter
+    def __init__(self, platform=sys.platform, *args, **kwargs):
```
sobolevn: This `platform`s should be defined as constants in https://github.com/lk-geimfari/mimesis/tree/master/mimesis/constants.

mimesis/providers/path.py
```diff
@@ -35,7 +34,7 @@ def home(self):
         :Example:
             /home/
         """
-        if sys.platform == 'win32':
+        if self.platform == 'win32':
```
sobolevn: I would suggest not to hardcode that.

mimesis/providers/path.py
```diff
@@ -87,5 +87,5 @@ def project_dir(self, user_gender='female'):
             /home/sherika/Development/Falcon/mercenary
         """
         project = self.random.choice(PROJECT_NAMES)
-        return os.path.join(
-            self.dev_dir(user_gender), project)
+        path = '\\' if self.platform == 'win32' else '/'
+        return (self.dev_dir(user_gender) + '{}' + project).format(path)
```
sobolevn: Why using `return ()`?

mimesis/providers/path.py
sobolevn: Maybe it is a good idea to create a structure with: 
1. name
2. path separator
3. other useful things

@lk-geimfari ?

mimesis/providers/path.py
```diff
@@ -9,10 +8,10 @@
 class Path(BaseProvider):
     """Class that provides methods and property for generate paths."""
 
-    def __init__(self, *args, **kwargs):
-        # TODO: platform should be a parameter
+    def __init__(self, platform=sys.platform, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.__p = Personal('en')
+        self.platform = platform
```
sobolevn: Could you please document all the platforms we support? What happens when user goes with `Path('does-not-exist')`?

sobolevn:
Good work!

mimesis/providers/path.py
lk-geimfari: @sobolevn Agree. @Valerievich Please, look at review.

Valerievich: @sobolevn what structure? can u please describe your vision more detail?

mimesis/providers/path.py
Valerievich: ok

mimesis/providers/path.py
Valerievich: Deal

mimesis/providers/path.py
Valerievich: Because 'project_dir()' already had this instruction. Is it wrong?

sobolevn: Oh, now I see. It is not `return (...)`, it is some complex expression hidden there. 
It is better to reduce the complexity here.

mimesis/providers/path.py
Valerievich: Do you mean constant.py should contains smth like:
```
{
    'WINDOWS': {'name': 'win32',
                'path_separator': '\\'},
    'LINUX2': {'name': 'linux2',
               'path_separator': '/'},
}
```

Valerievich: I have some idea about that structure and `Path()`. Just wait for it please.

sobolevn: And `'WINDOWS'` should be a constant as well.

mimesis/providers/path.py
Valerievich: @lk-geimfari BTW, which doc.file should I update? `mimesis/docs/guide.rst` ?

sobolevn: Please, update doc-strings also.

Valerievich:
Didn't get why `Codecov/path` were not successful

PR #204, opened 2017-08-22
Add Datetime.timezone

k4nar:
This should be pretty much all of them. I've extracted them from https://timezonedb.com/time-zones.


mimesis/providers/date.py
```diff
@@ -2,6 +2,7 @@
 from calendar import monthrange
 
 from mimesis.data import ROMAN_NUMS
+from mimesis.data import TIMEZONES
```
lk-geimfari: Please, use following format:
```
from mimesis.data import ROMAN_NUMS, TIMEZONES
```

lk-geimfari:
@k4nar Thank you! Please, look at very small change request.

lk-geimfari:
@k4nar What do you think about GMT offset. Can it be useful? I mean following: 

```python
>>> Datetime().gmt_offset()
'UTC +02:00'
```

lk-geimfari:
I'll fix it by myself, thank you.

k4nar:
> I'll fix it by myself, thank you.

Ok great, I was about to do it :) .

> What do you think about GMT offset. Can it be useful?

I don't remember having this need before, but some might find it useful. However for a random timezone-aware datetime, I would advise to generate a datetime in UTC and switch to a random timezone with an appropriate library.

PR #223, opened 2017-09-07
Fix for generate new documentation

Valerievich:
- updated source file `api.rst` for documentation
- resolved WARNINGs `make html` command

So now we can exec. `make html` and get a relevant documentation.


sobolevn:
@lk-geimfari is readme bundled with docs right now?

lk-geimfari:
@sobolevn Seems like not bundled.

Issue #96, opened 2017-01-07
Check correctness of all data for all locales.


lk-geimfari:
We have the support of 33 languages and it would be great if native-speakers of one's will check the correctness of data for his own language.

For example. I'm Russian and I'm sure of the correctness of the data for this language. But we also want to be sure of the correctness of German (`de`, `de-ch`), Italian (`it`) and other languages.

Checked locales:
- [ ] `cs`
- [ ] `el`
- [ ] `es-mx`
- [ ] `es`
- [ ] `et`
- [ ] `fa`
- [ ] `fi`
- [ ] `hu`
- [ ] `is`
- [ ] `ja`
- [ ] `kk`
- [ ] `ko`
- [ ] `nl-be`
- [ ] `nl`
- [ ] `no`
- [ ] `sv`
- [ ] `zh`

ZeFifi:
Hi @lk-geimfari ! I'm confused: could you tell me which files need to be checked? I could help with the french ones ;)

lk-geimfari:
@ZeFifi All files related to your locale, namely: `mimesis/data/fr/`

ZeFifi:
@lk-geimfari Ok cool, I'm gonna take a look at it, I saw some typos ;)

lk-geimfari:
@ZeFifi Thank you very much!

ZeFifi:
Everything looks OK now :) 

Please not that in the text.json file there's this sentence, in english: "This is the first age that's ever paid much attention to the future, which is a little ironic since we may not have one.". Would you like me to translate it or is it meant to remain in english since it's a quote?

lk-geimfari:
@ZeFifi In `fr` all data must be in French. If you can find more data in French then update it! It would be awesome!

ZeFifi:
Sure, I'm gonna recheck all the JSON ;)

ZeFifi:
@lk-geimfari It's OK for the french translations now ;) I submitted my last pull request! You can check the FR box ;) Have a great day, it was cool to contribute to your project!

lk-geimfari:
@ZeFifi Thanks!

ar4s:
@lk-geimfari I will check for pl

ChloeAnn:
Hi, I would love to contribute, I will check en-gb.

lk-geimfari:
@ChloeAnn Okay, thanks!

wbkoetsier:
[...]

PR #235, opened 2017-10-02
Fix state information in French and Spanish locale - see #96

sebasrp:
Fix state information in French and Spanish locale - see #96

Normalized the abbreviation to use standard ISO_3166
* French info: https://fr.wikipedia.org/wiki/ISO_3166-2:FR
* Spanish info: https://es.wikipedia.org/wiki/ISO_3166-2:ES


sebasrp:
Not sure I understand why the code coverage would reduce and how to fix that. @lk-geimfari can you help me understand how to make the check pass?

lk-geimfari:
@sebasrp It's okay. All py.test tests passed. Thanks!

PR #236, opened 2017-10-03
Add values for PT-BR in datetime.json and correct some values in PT and PT-BR

marcosvafg:


mimesis/data/pt-br/datetime.json
```diff
@@ -1,2 +1,66 @@
 {
+	  "day": {
+    "name": [
```
lk-geimfari: It's not okay. Please, reformat the file using: http://jsoneditoronline.org

PR #247, opened 2017-10-04
Fixes some typos for Italian #96

verginer:
There where several typos.
  - Major city names were english (i.e. Rome, should be Roma) 
  - Double entries
  - and some other fixes


lk-geimfari:
@verginer Thanks!

Issue #250, opened 2017-10-12
Issue with telephone() of provider Personal()


lk-geimfari:
So, it's returns too many similar data. We need to fix it **immediately**.

This method is based on the `custom_code()` method of the `Code()` provider and it should works like `custom_code()`, but it doesn't.

How works `telephone()` now:

```python
>>> from mimesis import Personal

>>> s = {Personal('ru').telephone() for _ in range(2000)}
{'+7-(967)-072-06-28', '+7-(915)-272-71-92', '+7-(937)-072-06-28', ... }
>>> len(s)
14 # Only 14 unique values from 2000.
```
How it should work:

```python
>>> import random

>>> from mimesis import Code
>>> mask = random.choice([
    "+7-(96#)-###-##-##",
    "+7-(92#)-###-##-##",
    "+7-(93#)-###-##-##",
    "+7-(91#)-###-##-##",
    "+7-(999)-###-##-##",
    "+7-(9##)-###-##-##",
    "+7-(909)-###-##-##",
    "+7-(968)-###-##-##"
])

>>> c = {Code('ru').custom_code(mask=mask) for _ in range(2000)}
{'+7-(921)-395-52-94', '+7-(955)-222-89-60', '+7-(978)-727-56-77', ... }
>>> len(c)
2000 # All values unique.
```
Similar problems are observed with other methods, such as `identifier()`.

P.S As i can judge, the problem not in a `seed`.

simba3447:
May I start working on this one?

lk-geimfari:
@sammyshj Sure! You're assigned. Thanks!

lk-geimfari:
Fixed in #251

PR #251, opened 2017-10-12
Fixed custom Random class in helpers.py

simba3447:
PR for [#250](https://github.com/lk-geimfari/mimesis/issues/250)
Removed super class initialization in random class.

### DIFF
--- a/mimesis/helpers.py
+++ b/mimesis/helpers.py
@@ -6,7 +6,7 @@ class Random(random.Random):
     """Custom Random() class for the possibility of extending."""
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


mimesis/helpers.py
```diff
@@ -6,7 +6,7 @@ class Random(random.Random):
     """Custom Random() class for the possibility of extending."""
 
     def __init__(self):
```
lk-geimfari: Remove this method, please.

lk-geimfari:
@simba3447 Great work! Thank you!

PR #260, opened 2017-10-22
[WIP] Type hinting

lk-geimfari:
I have implemented type hints for module `utils.py`.


mimesis/utils.py
```diff
@@ -7,30 +7,35 @@
...
 from random import choice
 
-from mimesis.exceptions import UnexpectedGender, UnsupportedLocale
-from mimesis.settings import SUPPORTED_LOCALES
+from mimesis.exceptions import (
+    UnexpectedGender,
+    UnsupportedLocale,
+)
+import mimesis.settings as settings
+import mimesis.type as types
```
lk-geimfari: I have collected all types in this module and import ones from it. Is it good idea?

mimesis/type.py
```diff
@@ -0,0 +1,16 @@
+from typing import Dict, List, Union, Mapping, Any
```
sobolevn: I would also use the name `typing` for this module. `mimesis.typing`.

mimesis/utils.py
sobolevn: Yeap, it is!

sobolevn:
Dropping `python3.3.` and `python3.4` will be a new minor release (or even major!).

lk-geimfari:
@DefaltSimon you should use types from `mimesis.typing`. `utils.py` is great example. Thanks!

Issue #258, opened 2017-10-21
Type hinting


sobolevn:
I have started to use `mypy`. Aaaand it is not so good right now.

Libraries lack typing support. And I know that `mimesis` support only `python3`.
Do you consider adding type hints?

There are some advantages:
- stricter API
- static analysis support to improve quality

There are some disadvantages:
- type hinting was introduced in `python3.5`, so any version prior to that will have to use special comments or other hacks
- a lot of manual work
- possible API changes

lk-geimfari:
Do you mean annotations? 

Something like this:
```
def function(num: int) -> float:
    pass
```

Right?

sobolevn:
Yes.

lk-geimfari:
I thinks that we can add annotations. What about Python 3.4?

lk-geimfari:
[Seems](https://www.python.org/dev/peps/pep-3107/) like annotations was added in Python 3.0. And we we lose nothing.

sobolevn:
https://www.python.org/dev/peps/pep-0484/

lk-geimfari:
@sobolevn Soooo... What shall we do? Refusing to support Python 3.4 doesn't seem to be a good idea.

sobolevn:
It actually is in some cases. Since upgrading from `3.4` to `3.5` is painless in 90% of situations.

lk-geimfari:
@sobolevn Let's do this!

sobolevn:
Please, note that variable annotations is not supported in `3.5`. https://stackoverflow.com/questions/39971929/what-are-variable-annotations-in-python-3-6/39973133#39973133

lk-geimfari:
@sobolevn I will add annotations for functions and methods. Can i hope for your help with situations which confuse me?

lk-geimfari:
Example:

```python
    def randints(self, amount: int = None, a: int = 1, b: int = 100) -> list:
        """Generate list of random integers.

        :param amount: Amount of elements.
        :param a: Minimum value of range.
        :param b: Maximum value of range.
        :return: List of random integers.
        """

        if not amount:
            amount = 3

        return [self.randint(a, b)
                for _ in range(amount)]
```

lk-geimfari:
Is it right `amout: int = None`. I mean Is it should be integer default value?

lk-geimfari:
@sobolevn How to be if the function returns multiple types using? Should we use `typing` module?

lk-geimfari:
[...]

PR #267, opened 2017-10-26
Type hinting

lk-geimfari:
What was done:

- [x] Corrected most annotations.
- [x] Removed solutions added to support Python 3.3.
- [x] Added more types to `typing.py`.
- [x] Removed Python 3.3/3.4 from classifiers in `setup.py`.
- [x] Updated Python version badge.
- [x] Other minor changes and improvements.

Summary:

At this moment i have confused by few messages of `mypy`. I have created file `MYPY_REPORT.txt` and i hope for help with this issues.

Issue:  #258.


lk-geimfari:
@sobolevn I hope for your help.

MYPY_REPORT.txt
```diff
@@ -0,0 +1,3 @@
+mimesis/providers/units.py:299: error: Argument 1 to "choice" of "Random" has incompatible type "List[str]"; expected "Sequence[Dict[str, List[str]]]"
+mimesis/providers/generic.py:71: error: Item "function" of "Union[Callable[..., Any], Any]" has no attribute "Meta"
```
sobolevn: You are using a class, not function.

sobolevn: `Provider` should be a protocol: http://mypy.readthedocs.io/en/latest/class_basics.html#simple-user-defined-protocols

lk-geimfari: @sobolevn Should i create `protoclos.py` or just create it in `typing.py`?

sobolevn: IMHO stick to `typing.py`

lk-geimfari: @sobolevn Seems like i have not `typing_extensions` module in my version of Python 3.5.2 on Ubuntu 16.04:

sobolevn: https://github.com/python/typing/issues/467

lk-geimfari: So, it's mean that we should add third-party dependencies and i'm not sure that it's worth it. I will just ignore places where we need Protocol.

docs/conf.py
```diff
@@ -33,6 +33,7 @@
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx_autodoc_typehints'
```
lk-geimfari: Trailing comma.

bjd2385:
I'm going to propose the following stylistic guidelines while contributing type annotations.

- Follow PEP 8.
- Use `mimesis.typing` for defining Mimesis-only type variables, such as `TimeStamp`, `JSON`, etc. It's actually rather annoying to have to keep typing `types.Any`, when we should be importing them directly from Python's `typing` module at the top of the file, i.e. `from typing import Any`.
- For type checking I'm using the following command (are there any additional restrictions you'd like to impose?): `mypy --strict --strict-optional --check-untyped-defs`

Another thing I'm seeing over and over again is the use of spaces before default arguments, e.g. the following, which can be found in `helpers.py`:

> def randints(self, amount: int = 3, a: int = 1, b: int = 100) -> list:

This should instead be, per PEP 8 guidelines,

> def randints(self, amount: int =3, a: int =1, b: int =100) -> list:

I know it's a small thing to pick lol, it's just one of my pet peaves :P

lk-geimfari:
@bjd2385 I agree that spaces in arguments is not cool, but only when arguments have not annotations.  But in our case this is what PyCharm do automatically and i like it, because it's more readable.

```
def randints(self, amount: int = 3, a: int = 1, b: int = 100) -> list:
    ...
```

lk-geimfari:
@bjd2385 As for the rest of your comments, I agree with you and i will fix everything.

lk-geimfari:
@bjd2385 Also i recommend you look at this [PEP](https://www.python.org/dev/peps/pep-3107/#syntax).

PR #268, opened 2017-10-26
Checked en-gb and added data to science (chemical elements) and text …

ChloeAnn:
…(answers and colours)


lk-geimfari:
@ChloeAnn Hi! There no need to add science data, cuz it's inherit from locale `en`, like other data which is common for all countries where English is official language.

lk-geimfari:
@ChloeAnn I'll add remark to contributing guidelines. Thanks!

PR #272, opened 2017-10-29
setup.py encoding error

tamlt2704:
setup.py encoding error


lk-geimfari:
We don't support Python 2.7. Only 3.5 and higher.

PR #276, opened 2017-10-30
Some basic code/comment updates

bjd2385:
Just some basic updates I found while reviewing the code base.

By the way, this seems like a cool project!


lk-geimfari:
@bjd2385 Thanks!

PR #277, opened 2017-10-30
Some basic updates

bjd2385:
There's plenty more here to work on. As always, comments/criticism welcome :)

### DIFF
--- a/mimesis/decorators.py
+++ b/mimesis/decorators.py
@@ -1,4 +1,6 @@
-import functools
+from typing import Any, Callable, TypeVar, Generator, Iterable, cast
+U = TypeVar('U', covariant=True)
+
 from string import (
     ascii_letters as letters,
     digits,
@@ -7,23 +9,24 @@ from string import (
 
 from mimesis import data
 from mimesis.exceptions import UnsupportedLocale
-import mimesis.typing as types
+
+import functools
 
 
-def romanized(locale: str = '') -> types.Callable:
+def romanized(locale: str ='') -> Callable:
     """Romanization of the Cyrillic alphabet (transliterating the Cyrillic language
     from the Cyrillic script into the Latin alphabet).
 
-    .. note:: At this moment it's work only for `ru`, `uk`, `kk`.
+    .. note:: At this moment it works only for `ru`, `uk`, `kk`.
 
     :param locale: Function.
     :return: Latinized text.
     :rtype: types.Callable
     """
 
-    def romanized_deco(func):
+    def romanized_deco(func: Callable[..., U]) -> U:
         @functools.wraps(func)
-        def wrapper(*args, **kwargs):
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
             try:
                 alphabet = data.ROMANIZATION_DICT[locale]
                 # Add common cyrillic common letters
@@ -37,7 +40,7 @@ def romanized(locale: str = '') -> types.Callable:
                     'Locale {0} is not supported yet.'.format(locale),
                 )
             result = func(*args, **kwargs)
-            txt = ''.join([alphabet[i] for i in result if i in alphabet])
+            txt = ''.join(cast(Iterable, (alphabet[i] for i in result if i in alphabet)))
             return txt
 
         return wrapper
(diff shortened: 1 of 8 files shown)


mimesis/decorators.py
```diff
@@ -7,23 +9,24 @@
 
 from mimesis import data
 from mimesis.exceptions import UnsupportedLocale
-import mimesis.typing as types
 
+import functools
```
lk-geimfari: pep8, functools is in standard library.

mimesis/decorators.py
```diff
@@ -1,4 +1,6 @@
-import functools
+from typing import Any, Callable, TypeVar, Generator, Iterable, cast
+U = TypeVar('U', covariant=True)
```
lk-geimfari: What is mean `U`?

mimesis/providers/address.py
```diff
@@ -84,7 +87,7 @@ def address(self) -> str:
 
         )
 
-    def state(self, abbr: bool = False) -> str:
+    def state(self, abbr: bool =False) -> str:
```
lk-geimfari: https://www.python.org/dev/peps/pep-3107/#syntax

mimesis/providers/numbers.py
```diff
@@ -1,14 +1,15 @@
-import array
-import mimesis.typing as types
+from mimesis import typing as types
 
 from mimesis.providers.base import BaseProvider
 
+import array
```
lk-geimfari: 1. standard library imports
2. related third party imports
3. local application/library specific imports

mimesis/decorators.py
```diff
@@ -37,16 +40,16 @@ def wrapper(*args, **kwargs):
                     'Locale {0} is not supported yet.'.format(locale),
                 )
             result = func(*args, **kwargs)
-            txt = ''.join([alphabet[i] for i in result if i in alphabet])
+            txt = ''.join(cast(Iterable, (alphabet[i] for i in result if i in alphabet)))
```
bjd2385: I'm having a problem with `join` here, so this isn't yet in a working condition; e.g., I shouldn't have tried to `cast` this value to `Iterable`, it's just meant to show that `join` is complaining about requiring an `Iterable` type. Per `help(''.join)`:

> join(...) method of builtins.str instance
    S.join(iterable) -> str

mimesis/decorators.py
```diff
@@ -56,16 +59,16 @@ def type_to(new_type: types.Callable,
...
+        def wrapper(*args: Any, **kwargs: Any) -> U:
             result = func(*args, **kwargs)
             result = new_type(result)
 
             if check_len and len(result) == 1:
                 return result[0]
             return result
 
-        return wrapper
+        return cast(U, wrapper)
```
bjd2385: I'm still working out how to annotate these decorators properly :P

mimesis/helpers.py
```diff
@@ -18,10 +20,9 @@ def randints(self, amount: int = 3, a: int = 1, b: int = 100) -> list:
         if not amount:
```
bjd2385: I feel like we should restrict this further now that I'm looking at it; e.g., the predicate could probably be `amount == 0`, and in an `else` clause `raise ValueError('Amount out of range')` or something.

mimesis/providers/numbers.py
bjd2385: I'll follow that convention from here on out then.

mimesis/providers/address.py
bjd2385: If that's the convention you'd like to see in this code base, then that's what I'll follow.

mimesis/decorators.py
bjd2385: @lk-geimfari It's a type variable. I've already updated this too in a commit i'll make later with fixes as to your other comments here.

lk-geimfari:
@bjd2385 I suggest pull latest changes from `type-hinting` branch.

bjd2385:
@lk-geimfari Will do. As you can probably tell, I'm still learning about version control and GitHub's workings. Thanks for being so patient with me :)

lk-geimfari:
@bjd2385 Thank you for the contributing!

lk-geimfari:
@bjd2385 Ping.

PR #278, opened 2017-10-31
Fix sphinx warnings

py-crash:
Different warnings were fixed for the sphinx documentations. They were
mostly underline oder overline too shorts as well as missing indedations
either in the documentation files or in the docstings

The only one missing is 'Duplicate explicit target name: "here".' in the
index.rst file. I was unchanged since here refences two different
documents.


py-crash:
I found some problems while using theese docs as a base for writing the docs for lk-geimfari/wasserstoff#6, so i decided to git them a fix.

lk-geimfari:
@py-crash Thank you very much for fix this. Great job!

PR #282, opened 2017-11-11
Fixed Numbers.primes

### LINKED ISSUE #280: Numbers.primes returns odd numbers, not primes
The logic for primes in `Numbers.primes` is incorrect, and is instead for odd numbers:
```py
nums = array.array('L', (i for i in range(start, end) if i % 2))
return nums.tolist() if to_list else nums
```

faheel:
Closes #280


faheel:
I removed the `to_list` argument from `primes` because all the computations were done in lists as I found lists to be faster than arrays for computation.

lk-geimfari:
@faheel It's okay!

mimesis/providers/numbers.py
```diff
@@ -34,24 +37,40 @@ def floats(self, n: int = 2, type_code: str = 'f',
...
-        nums = array.array('L', (i for i in range(start, end) if i % 2))
-        return nums.tolist() if to_list else nums
+        global initial_primes
+
+        # if request is inside initial_primes return a subset of it.
+        if (end <= initial_primes[-1]):
+            return [p for p in initial_primes if start <= p <= end]
+
+        # remove case when start == 1
+        if (start < 2):
```
lk-geimfari: We don't use redounded parentheses in `if` statement. [PEP8](https://www.python.org/dev/peps/pep-0008/).

mimesis/providers/numbers.py
```diff
@@ -34,24 +37,40 @@ def floats(self, n: int = 2, type_code: str = 'f',
...
+
+        :return: A list of prime numbers from start to end.
+        :rtype: list
         """
-        nums = array.array('L', (i for i in range(start, end) if i % 2))
-        return nums.tolist() if to_list else nums
+        global initial_primes
+
+        # if request is inside initial_primes return a subset of it.
+        if (end <= initial_primes[-1]):
```
lk-geimfari: Look at the comment below.

mimesis/providers/numbers.py
```diff
@@ -34,24 +37,40 @@ def floats(self, n: int = 2, type_code: str = 'f',
...
+        for p in Numbers.primes(1, int(end**0.5) + 1):
+            s = max(p**2, start + (p - start % p) % p)
+            for i in range(s - start, end - start, p):
+                sieve[i] = False
+
+        # create a list of primes from the sieve
+        primes_list = [i + start for i, x in enumerate(sieve) if x]
+
+        # update the list of initial primes when bigger one is available
+        if (start <= 2 and primes_list > initial_primes):
```
lk-geimfari: Remove redounded parentheses, please.

mimesis/providers/numbers.py
```diff
@@ -4,6 +4,9 @@
 from mimesis.typing import Array, Number
 
 
+initial_primes = [2]
```
lk-geimfari: That is not better idea.

mimesis/providers/numbers.py
```diff
@@ -34,24 +37,40 @@ def floats(self, n: int = 2, type_code: str = 'f',
...
-        :param bool to_list: Convert array to list.
-        :return: An array of floating-point numbers.
-        :rtype: Array
+
+        :return: A list of prime numbers from start to end.
+        :rtype: list
         """
-        nums = array.array('L', (i for i in range(start, end) if i % 2))
-        return nums.tolist() if to_list else nums
+        global initial_primes
```
sobolevn: Nope, `global` is not allowed.

lk-geimfari:
@faheel Just look at this [example](https://github.com/keon/algorithms/blob/master/math/primes_sieve_of_eratosthenes.py), please.  It's easy and much readable. If you can modify it with `start` and `end` we can add it.

mimesis/providers/numbers.py
sobolevn: We don't really need brackets here.

mimesis/providers/numbers.py
sobolevn: And here too.

mimesis/providers/numbers.py
sobolevn: And here too.

mimesis/providers/numbers.py
```diff
@@ -34,24 +37,40 @@ def floats(self, n: int = 2, type_code: str = 'f',
...
+        if (end <= initial_primes[-1]):
+            return [p for p in initial_primes if start <= p <= end]
+
+        # remove case when start == 1
+        if (start < 2):
+            start = 2
+
+        # create a sieve and remove all non-primes from it
+        sieve = [True] * (end - start)
+        for p in Numbers.primes(1, int(end**0.5) + 1):
```
sobolevn: I would love to see some comments about what going on here.
A link to some algorithm explanation would also be very helpful.

tests/test_data/test_numbers.py
```diff
@@ -14,11 +14,11 @@ def test_floats(numbers):
 
 def test_primes(numbers):
     result = numbers.primes()
-    assert len(result) == 499
-    assert isinstance(result, array.array)
+    assert len(result) == 168
```
sobolevn: This number seems like magic to me. Could you please explain what happens here?
A comment to some papers or articles would be awesome.

tests/test_data/test_numbers.py
```diff
@@ -14,11 +14,11 @@ def test_floats(numbers):
...
+    assert isinstance(result, list)
 
-    result = numbers.primes(to_list=True)
-    assert len(result) == 499
+    result = numbers.primes(500, 500000)
+    assert len(result) == 41443
```
sobolevn: The same goes here.

tests/test_data/test_numbers.py
faheel: Instead of checking whether every number in the returned list is a prime, this simply checks that the size of the list = number of primes in the given range (which is 1-999 by default).

PR #283, opened 2017-11-12
russian jsc corrected

mfperminov:
SInce about 2014 there is no such type of business entity like "Закрытое акционерное общество" ("Closed joint-stock company"). Correct version "Акционерное общество" ("Joint-stock company").


lk-geimfari:
@mfperminov Thanks!

PR #289, opened 2017-11-16
[WIP] Enum args

lk-geimfari:
**Added**:

- Added `Field` for generating data by schema
- Added `category_of_website` and `port` to `Internet` data provider
- Added new module `enums.py` and support of enums in arguments of methods
- Added `dna`, `rna` and `atomic_number` to `Science` data provider
- Added `schoice` method for `Random`
- Added alias `last_name` for `surname` in `Personal` data provider
- Added alias `province`, `region`, `federal_subject` for `state` in `Address` data provider
- Added annotations for all methods and functions for supporting type hints
- Added module `typing.py` for custom types
- Added new data provider `Payment`
- Added new methods to `Payment`: `credit_card_network`, `bitcoin_addres`, `credit_card_owner`

**Fixed**:
- Fixed issue with `primes` in `Numbers` data provider
- Fixed issue with repeated output on using `Code().custom code`
- Other minor fix and improvements

**Mover/Removed**:
- Moved `credit_card`, `credit_card_expiration_date`, `cid`, `cvv`, `paypal` and `bitcoin` to `Payment` from `Personal`
- Moved `custom_code` to `utils.py` from `providers.code.Code`
- Removed some useless methods
- Removed module `constants`, in view of adding more convenient and useful module `enums`
- Removed non informative custom exception `WrongArgument` and replaced one with `KeyError` and `NonEnumerableError`

**Updated/Renamed**:
- Updated data for `de-at`, `en`, `fr`, `pl`, `pt-br`, `pt`, `ru`, `uk`
- Other minor updates in other languages


mimesis/enums.py
```diff
@@ -0,0 +1,23 @@
...
+    DEFAULT = (1, 65535)
+    WELL_KNOWN = (1, 1023)
+    EPHEMERAL = (49152, 65535)
+    REGISTERED = (1024, 49151)
+
+
+class Gender(Enum):
+    FEMALE = 'female'
+    MALE = 'male'
+    F = FEMALE
```
lk-geimfari: I'm not sure that is good idea, but short alias can be useful sometime.

mimesis/providers/personal.py
```diff
@@ -99,49 +96,52 @@ def surname(self, gender: Gender = 0) -> str:
...
+        """An alias of self.surname.
+
+        :param gender: Gender's enum object.
+        :return: Last name.
+        """
         return self.surname(gender)
 
-    def title(self, gender: Gender = 0,
-              title_type: str = 'typical') -> str:
+    def title(self, gender: Gender = Gender.RANDOM,
```
lk-geimfari: Or maybe `Gender.ANY`?

mimesis/enums.py
sobolevn: No, I don't think so. There is autocomplete for this kind of things.
But for me it would not be clear: is `F` strictly the same as `FEMALE`?

mimesis/providers/personal.py
sobolevn: `RANDOM` seems the best choice to be. As it clearly states how the gender is selected. How `ANY` gender is selected?

mimesis/enums.py
```diff
@@ -0,0 +1,23 @@
...
+    M = MALE
+    RANDOM = choice([FEMALE, MALE])
```
sobolevn: This is broken, since it selected the gender only once when created.

```
>>> class Gender(Enum):
...    MALE = 'male'
...    FEMALE = 'female'
...    RANDOM = choice([MALE, FEMALE])
... 
>>> Gender.RANDOM
<Gender.MALE: 'male'>
>>> Gender.RANDOM
<Gender.MALE: 'male'>
>>> Gender.RANDOM
<Gender.MALE: 'male'>
>>> Gender.RANDOM
<Gender.MALE: 'male'>
>>> Gender.RANDOM
<Gender.MALE: 'male'>
```

mimesis/enums.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class TitleType(Enum):
+    TYPICAL = 'typical'
+    ACADEMIC = 'academic'
+    RANDOM = choice([TYPICAL, ACADEMIC])
```
sobolevn: The same here.

mimesis/enums.py
```diff
@@ -0,0 +1,23 @@
+from enum import Enum
+from random import choice
+
+
+class PortRange(Enum):
+    DEFAULT = (1, 65535)
```
sobolevn: `ALL`?

mimesis/enums.py
lk-geimfari: @sobolevn I'll remove ones.

mimesis/enums.py
lk-geimfari: @sobolevn Do you have any suggestions, how to get random value each time?

mimesis/enums.py
lk-geimfari: @sobolevn Sure!

mimesis/enums.py
sobolevn: Maybe we can try to use `lambda` for that?

lk-geimfari: @sobolevn I have tried this, but it's broke type hints and looks not so good, although it works:

```python
class TitleType(Enum):
    # ...
    RANDOM = lambda: choice(['typical', 'academic'])
```

```python
for i in range(5):
    print(TitleType.RANDOM())

'typical'
'academic'
'typical'
'academic'
'academic'
```

lk-geimfari: We can just return string `random` and choice random value in method:

```python
  if title_type and title_type in TitleType:
            if title_type == TitleType.RANDOM:
                title_type = self.random.choice(
                    [TitleType.ACADEMIC, TitleType.TYPICAL]
                )
            titles = titles[title_type.value]
```
Here i'm not sure too.

lk-geimfari: @sobolevn Do you have any better solution in mind?

lk-geimfari: So, i have created question on [SOF](https://stackoverflow.com/questions/47353555/how-to-get-random-value-of-attribute-of-enum-on-each-iteration).

lk-geimfari: @sobolevn I like this [solution](https://stackoverflow.com/a/47354435/5461570). What do you think about one?

sobolevn: I like this one:

```python
import random
import enum
class RANDOM_ATTR(enum.EnumMeta):
    @property
    def RANDOM(self):
        return random.choice(list(self))


class Gender(enum.Enum,metaclass=RANDOM_ATTR): #this syntax works for python3 only
    FEMALE = 'female'
    MALE = 'male'


print(Gender.RANDOM)
```

Results:

```
>>> print(Gender.RANDOM)
Gender.MALE
>>> print(Gender.RANDOM)
Gender.MALE
>>> print(Gender.RANDOM)
Gender.FEMALE
>>> print(Gender.RANDOM)
Gender.FEMALE
>>> print(Gender.RANDOM)
Gender.FEMALE
```

sobolevn: It could even be a simple mixin, not a metaclass.

lk-geimfari: The problem with this solution  is that we have to many `enums` where we need random choice. We can't create metaclasses for each ones, it will looks dirty.

lk-geimfari: We can inherit this class:
```python
class Enum(enum.Enum, metaclass=enum.EnumMeta):

    RANDOM = None
    @classmethod
    def get_random(self):
        return choice(list(self))
```
Use it like that:
```python
def check_gender(gender: Gender = Gender.RANDOM) -> Gender:
    if not gender:
        return gender.get_random()

    elif gender and gender in Gender:
        return gender
    else:
        raise UnexpectedGender(
            'You should use "Gender" from the module mimesis.enums',
        )
```

lk-geimfari: I think that this is a good compromise.

lk-geimfari: Seems like my solution does not work correctly, sorry.

lk-geimfari: I take my words back.

lk-geimfari: @sobolevn It's can create some problems with type-hints. Should we use annotation like this `Union[Gender, property]`? Because `RANDOM` is property.

lk-geimfari: @sobolevn That's why i ask this question:
image

sobolevn: Oh, I was not aware of that. That's a problem. https://github.com/python/mypy/issues/220

sobolevn:
All in all. 

It think that putting `RANDOM` in `enum` will not going to work for multiple reasons:
1. that's not what `enum` is for
2. it is complicated to implement anyway

What do you came up with?

```python
class Gender(Enum):
    FEMALE = 'female'
    MALE = 'male'


def random_option(enum: Enum):  # I am not sure about return type
    return random.choice(list(self))
```

mimesis/enums.py
lk-geimfari: @sobolevn I don't quite understand how it works, but this descriptor fix this issue:

```python
class ClassPropertyDescriptor(object):
    """Author of this class is Mahmoud Abdelkader
    """
    def __init__(self, fget, fset=None):
        self.fget = fget
        self.fset = fset

    def __get__(self, obj, klass=None):
        if klass is None:
            klass = type(obj)
        return self.fget.__get__(obj, klass)()

    def __set__(self, obj, value):
        if not self.fset:
            raise AttributeError("Can't set attribute")
        type_ = type(obj)
        return self.fset.__get__(obj, type_)(value)

    def setter(self, func):
        if not isinstance(func, (classmethod, staticmethod)):
            func = classmethod(func)
        self.fset = func
        return self


def classproperty(func):
    if not isinstance(func, (classmethod, staticmethod)):
        func = classmethod(func)
    return ClassPropertyDescriptor(func)
```

lk-geimfari:
@sobolevn I have checked it, it's works fine.

mimesis/enums.py
```diff
@@ -0,0 +1,115 @@
+from enum import Enum, EnumMeta
+from random import choice
+
+
+class AllowRandom(EnumMeta):
+    """This metaclass allow method get_random_item() which
+    equal to random field of enum for his subclasses.
+    """
+
+    def get_random_item(self):
```
lk-geimfari: @sobolevn At this moment we have some issues related to this method. It's because we don't annotated this method:
```
mimesis/providers/units.py:325: error: Item "None" of "Optional[PrefixSign]" has no attribute "value"
mimesis/providers/cryptographic.py:36: error: Item "None" of "Optional[Algorithm]" has no attribute "value"
mimesis/providers/cryptographic.py:37: error: Item "None" of "Optional[Algorithm]" has no attribute "value"
mimesis/providers/code.py:61: error: Incompatible types in assignment (expression has type "str", variable has type "Optional[ISBNFormat]")
mimesis/providers/code.py:61: error: Item "None" of "Optional[ISBNFormat]" has no attribute "value"
# ...
# ...
```
How can we indicate in the annotations that the method returns self?

I try this:

```
def get_random_item(self) -> Enum:
    ...
```

But it did not help.

mimesis/providers/personal.py
```diff
@@ -272,23 +273,25 @@ def email(self, domains: Union[tuple, list] = None) -> str:
...
-        ]
-        url = 'http://' + self.random.choice(urls)
-        username = self.username(template='U_d')
+        if site in SocialNetwork:
+            site = SOCIAL_NETWORKS[site.value]
+        else:
+            raise NonEnumerableError(SocialNetwork)
+
+        url = 'https://www.' + site
+        template = self.random.choice(['l-d', 'l_d', 'Ud', 'ld'])
```
sobolevn: templates are doubled, you already have them somewhere else

tests/test_data/test_personal.py
```diff
@@ -133,8 +133,16 @@ def test_favorite_music_genre(_personal):
     assert result in MUSIC_GENRE
 
 
-def test_social_media_profile(_personal):
-    result = _personal.social_media_profile()
+@pytest.mark.parametrize(
+    'site', [
```
sobolevn: you are missing `None` as a param

lk-geimfari: @sobolevn Fixed in 6d2c7b9.

mimesis/providers/personal.py
lk-geimfari: @sobolevn I'll update it.

mimesis/enums.py
lk-geimfari: @sobolevn I really need your help here. I don't know how to be with this problem.

PR #296, opened 2017-11-23
builtin specprovider for locale nl

jorisdevrede:
Added the NLSpecProvider for the locale nl to generate a _burgerservicenummer_ or BSN. This is the Dutch equivalent of a social security number.

### DIFF
--- a/mimesis/builtins/nl.py
+++ b/mimesis/builtins/nl.py
@@ -5,7 +5,7 @@ class NetherlandsSpecProvider(BaseSpecProvider):
     """Provides data specific to nl locale"""
 
     class Meta:
-        name = 'nl_provider'
+        name = 'netherlands_provider'
 
     def bsn(self) -> str:
         """Generate a random, but valid burgerservicenummer (BSN).
@@ -17,7 +17,7 @@ class NetherlandsSpecProvider(BaseSpecProvider):
         """
 
         # a bsn is valid when its 9 digits pass the '11 test'
-        def _is_valid_bsn(number):
+        def _is_valid_bsn(number: str) -> str:
             total = 0
             multiplier = 9
 
--- a/mimesis/builtins/nl.py
+++ b/mimesis/builtins/nl.py
@@ -1,0 +1,35 @@
+from mimesis.builtins.base import BaseSpecProvider
+
+
+class NLSpecProvider(BaseSpecProvider):
+    """Provides data specific to nl locale"""
+
+    class Meta:
+        name = 'nl_provider'
+
+    def bsn(self) -> str:
+        """Generate a random, but valid burgerservicenummer (BSN).
+
+        :returns: random BSN
+
+        :Example:
+            255159705
+        """
+
+        # a bsn is valid when its 9 digits pass the '11 test'
+        def is_valid_bsn(number):
+            total = 0
+            multiplier = 9
+
+            for char in number:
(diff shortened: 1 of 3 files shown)


mimesis/builtins/nl.py
```diff
@@ -0,0 +1,35 @@
...
+        """Generate a random, but valid burgerservicenummer (BSN).
+
+        :returns: random BSN
+
+        :Example:
+            255159705
+        """
+
+        # a bsn is valid when its 9 digits pass the '11 test'
+        def is_valid_bsn(number):
```
lk-geimfari: This method must be private: `__is_valid_bsn`

mimesis/builtins/nl.py
```diff
@@ -0,0 +1,35 @@
+from mimesis.builtins.base import BaseSpecProvider
+
+
+class NLSpecProvider(BaseSpecProvider):
```
lk-geimfari: Maybe `NetherlandsSpecProvider`? It's much clearer for people which want import spec-provider for Netherlands.

lk-geimfari:
@jorisdevrede I have two notes, but generally this is nice! Please, look at the review.

mimesis/builtins/nl.py
jorisdevrede: You're right, although I believe this should be a single leading underscore `_is_valid_bsn` for non-public functions, right?

mimesis/builtins/nl.py
jorisdevrede: Yes. I will change it.

mimesis/builtins/nl.py
lk-geimfari: @jorisdevrede Absolutely.

mimesis/builtins/nl.py
lk-geimfari: @jorisdevrede Okay! Thanks. Let me know, please.

mimesis/builtins/nl.py
jorisdevrede: Done :-)

mimesis/builtins/nl.py
```diff
@@ -0,0 +1,35 @@
...
+        """Generate a random, but valid burgerservicenummer (BSN).
+
+        :returns: random BSN
+
+        :Example:
+            255159705
+        """
+
+        # a bsn is valid when its 9 digits pass the '11 test'
+        def _is_valid_bsn(number):
```
lk-geimfari: Annotation, please.

mimesis/builtins/nl.py
```diff
@@ -0,0 +1,35 @@
+from mimesis.builtins.base import BaseSpecProvider
+
+
+class NetherlandsSpecProvider(BaseSpecProvider):
+    """Provides data specific to nl locale"""
+
+    class Meta:
+        name = 'nl_provider'
```
lk-geimfari: `netherlands_provider`

mimesis/builtins/nl.py
jorisdevrede: Yes, of course. I've added annotations.

mimesis/builtins/nl.py
jorisdevrede: Good catch! I changed it and added an additional test to cover it properly.

PR #320, opened 2017-12-13
Renamed internal attribute self.data

lk-geimfari:
The library was had different attribute names, in which the data is stored (`self.data` and `self._data`). This inconsistency was from the very beginning of the library, ie some attributes of providers were called `self.data`, but other ones `self._data`.


sobolevn:
When you say that something was public and then you made it protected - it means that there are changes in the public API.

I don't see any other issues.

lk-geimfari:
@sobolevn Sorry, I incorrectly put it. I meant in the terms of backward compatibility.

lk-geimfari:
@sobolevn Also, I want to ask you what is a better name for the class which will represent any provider?

I have named it `Abstract`, but I'm not sure:
```python

_ = Abstract('en')
_('surname', gender=Gender.MALE)
```

PR #324, opened 2017-12-17
Seed support for providers

duckyou:


lk-geimfari:
@sobolevn Did you not add support for `seed`?

lk-geimfari:
@duckyou We have already added support of `seed`. There no need copy seed to all providers, because it's can be defined in BaseProvider which is the superclass for all data providers.

@sobolevn Correct me if I'm wrong.

mimesis/providers/address.py
```diff
@@ -132,7 +132,7 @@ def postal_code(self) -> str:
         """
 
         mask = self._data['postal_code_fmt']
-        return custom_code(mask=mask)
+        return custom_code(mask=mask, rnd=self.random)
```
lk-geimfari: It' better to move `custom_code` to  `CustomCodeMixin` and inherit it in `BaseProvider`.

mimesis/providers/file.py
```diff
@@ -10,9 +10,9 @@
 class File(BaseProvider):
     """Class for generate fake data for files."""
 
-    def __init__(self):
-        super().__init__()
-        self.__text = Text('en')
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.__text = Text('en', seed=self.seed)
```
lk-geimfari: It's internal attribute and developer should not manage it.

mimesis/providers/generic.py
```diff
@@ -38,17 +40,18 @@ def __init__(self, *args, **kwargs):
         self._science = Science
         self._code = Code
         self._transport = Transport
-        self.unit_system = UnitSystem()
```
lk-geimfari: The attributes starting with `_` is attributes which should be initialized lazily because they need locales for correct work. Other providers do not need locale for correct working because they contain international data.

duckyou:
@lk-geimfari I'll try to explain why I did so

> @duckyou We have already added support of seed. There no need copy seed to all providers, because it's can be defined in BaseProvider which is the superclass for all data providers.

... but some child provider didn't use inherited `random`.  For example:
```python
>>> from mimesis import Address
>>> a = Address(seed=42)
>>> a.postal_code()
'59080'
>>> a.postal_code()
'49715'
>>> a.postal_code()
'19811'
>>> a = Address(seed=42)
>>> a.postal_code()
'57034'
>>> a.postal_code()
'75790'
```
because `postal_code` call external function `custom_code`, where used new random object.

First commint https://github.com/lk-geimfari/mimesis/pull/324/commits/a759f17012cbb1365235cf72205596784ee826e8 fixed it.

Then i just write tests and fixes for pop-up issues :innocent:

> I can't merge this until because there are too changes.

Yeah... sorry for that. I should have firstly create issue post :worried:

mimesis/providers/generic.py
duckyou: `File` provider not lazy loaded, but when initialized it create new `Text` provider which use locales.

lk-geimfari:
@duckyou  I understand that you have done a big job, but first, we need to try the method I described earlier, i.e moving `custom_code` to `BaseProvider`.

Also seems like you have found a bug in our code and seed is really does not work.

I have created issue #325.

Issue #325, opened 2017-12-18
Seed does not work for some providers


lk-geimfari:
**Description:**
Seems like @duckyou have found a bug in our code and `seed` is really does not work sometime and we should fix it.

**Reason:**
It's happening because we use `custom_code` which use another `random` object.

**Decision:**
We should move utils which use `random` to `helpers.Random`

duckyou:
Thanks for the quick response, @lk-geimfari :smile:

I can try moving `custom_code` to `BaseProvider` like you said in #324  discussion

lk-geimfari:
@duckyou You know, it's better to move it to `helpers.Random`. There's just everything related to random selection.

lk-geimfari:
@duckyou after that, we can just call `self.random.custom_code`. Looks good.

lk-geimfari:
@duckyou If you don't want to do it, I can do it by myself, but if you want work on it, then, please, create new PR, because old is really hard to read :smile: . Thanks!

duckyou:
@lk-geimfari Hmm... You goddamn right!
That's may be simplest solution :relaxed:

I do this ;)

lk-geimfari:
@duckyou Thank you!

duckyou:
@lk-geimfari
https://github.com/lk-geimfari/mimesis/blob/af9c877d9bde654dff2f8be524b81a35460f2307/mimesis/enums.py#L6-L12

Method get_random_item also affected :worried:

lk-geimfari:
@duckyou We will not change it, because it's not really about `data`, but about `arguments` for function.

duckyou:
@lk-geimfari Issue explain:

```python
>>> from mimesis import Address
>>> from mimesis.enums import CountryCode
>>> a = Address(seed=42)
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'NIC'
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'BRN'
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'ARM'
>>> a = Address(seed=42)
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'NIC'
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'BRN'
>>> a.country_iso_code(fmt=CountryCode.ISO3)
'ARM'
...
```
This worked because we set format directly
... but if we omit this parameter

```python
...
>>> a = Address(seed=42)
>>> a.country_iso_code()
[...]

PR #327, opened 2017-12-18
Some fixes for issue #325

duckyou:
https://github.com/lk-geimfari/mimesis/issues/325#issuecomment-352356998

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -49,6 +49,7 @@
 - axcel [(axce1)](https://github.com/axce1)
 - Ruslan Valerievich [(Valerievich)](https://github.com/Valerievich)
 - Simon [(DefaltSimon)](https://github.com/DefaltSimon)
+- dy [(duckyou)](https://github.com/duckyou)
 
 ### Design
  - The illustration is taken from internet. If you know the creator of this octopus, then let us know.
--- a/mimesis/builtins/base.py
+++ b/mimesis/builtins/base.py
@@ -1,10 +1,7 @@
-from random import Random
-
+from mimesis.helpers import Random
 from mimesis.providers.base import ValidateEnumMixin
-from mimesis.utils import custom_code
 
 
 class BaseSpecProvider(ValidateEnumMixin):
     def __init__(self):
         self.random = Random()
-        self.code = custom_code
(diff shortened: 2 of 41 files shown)


lk-geimfari:
@duckyou I'll do a full review a little later and if it's okay - I'll merge it.

P.S Don't forget to add yourself to `CONTRIBUTORS.md`.

lk-geimfari:
@duckyou So, there one big problem. I think that testing all methods with `seed` is too painful and hard to maintain. We cant write two different tests for one method every time when we create new methods. I suggest removing this tests. We can test `seed` only using `BaseProvider`, i.e in file `test_base.py` if seed works for it, it will mean that `seed` works for all subclasses of `BaseProvider`:

```python
@pytest.fixture
def seeded_base():
    return BaseProvider(seed=42)

def test_base_with_seed(seeded_base):
    result = seeded_base.random.randint(1, 10)
    assert result == 2
```

**What should be done:** Remove all the tests which were added and replace them with the code above.

lk-geimfari:
@duckyou You are probably already tired of making corrections, but we must support the simplicity of the project and its tests. If you do not want to do this, then I understand you and I will do it myself. Thanks!

duckyou:
@lk-geimfari 

> @duckyou So, there one big problem. I think that testing all methods with seed is too painful and hard to maintain. We cant write two different tests for one method every time when we create new methods. I suggest removing this tests. We can test seed only using BaseProvider.

I partly agree. Two different tests for one method is awful design... my bad =(
But write tests only for BaseProvider also bad decision. Heres eg:
https://github.com/lk-geimfari/mimesis/pull/327/commits/53e19b8db4eda89d7c595dbc568b9566243ec2f1#diff-fb2e53700ab7e63642efe387f905686a
```python
    def words(self, quantity: int = 5) -> list:
        # Here is issue... some times normal words returnes in random order
        words = self._data['words'].get('normal')
        # Heres fix... when seed seted, its sorted words
        if self.seed: 
            words = sorted(words)
        words_list = [self.random.choice(words) for _ in range(quantity)]
        return words_list
```

IMHO Seed tests should be for each method. I recommend merge seed tests with its basic.

lk-geimfari:
@duckyou So, the situation with the previous PR was repeated here (not without my help). I'll create a new branch and applied your changes in this branch, but partially to better understanding what is happens.

lk-geimfari:
@duckyou I'll do it tomorrow, but for now, we'll leave it as it at that. Do not change anything, please.

lk-geimfari:
This PR was merged partially, as I said in our discussion.

PR #332, opened 2017-12-21
Fixed #325

### LINKED ISSUE #325: Seed does not work for some providers

lk-geimfari:


mimesis/helpers.py
```diff
@@ -74,3 +76,35 @@ def random_int(a: int, b: int) -> int:
...
+
+def get_random_item(enum: Any, seed: Optional[int] = None) -> Any:
+    """Get random item of enum object.
+
+    :param enum: Enum object.
+    :param seed: Seed for random.
+    :return: Random item of enum.
+    """
+    if seed is not None:
+        random.seed(seed)
```
duckyou: @lk-geimfari heres we have some issue. When we call this function with specific seed, it **always** be return same value
Why we cant just pass random object instance?

lk-geimfari: @duckyou We can do it. I just thought that we can do it with a minimum of changes. Create a branch from this branch with your changes, please.

PR #344, opened 2018-01-04
Added benchmarks for tests

lk-geimfari:


duckyou:
a little bit later i push commit which update setup.py file to ignore benchmarks (needed add `--benchmark-skip` argumet to `py.test`)

PR #350, opened 2018-01-07
Adds Pipfile, closes #348

### LINKED ISSUE #348: Replace requirements.txt with Pipfile
Pipenv (which works with Pipfile) is officially recommended Python packaging tool from Python.org. We should use it instead `requirements.txt `.

sobolevn:
Changes:
1. Removed all *.txt requirements
2. Added Pipfile and Pipfile.lock
3. Added docs/requirements.txt for readthedocs, since Pipfile is not
supported yet
4. Changed how CI installs requirements
5. Changed configuration files from multiple *.ini to a single setup.cfg
6. Fixed style issue in test
7. Removed `setup.py test` as it is not supported anymore


sobolevn:
@lk-geimfari can you please help me with `appveyor`? I don't know what's wrong, I don't have an access to windows machine (lucky me!).

lk-geimfari:
@sobolevn Seems like this command fix that problem: `pip3 install pew`. But I also work on Linux machine and I'm not sure how to fix this issue if command above is not working.

lk-geimfari:
@duckyou Do you work on Windows machine? Please, help us understand what is happening.

lk-geimfari:
@sobolevn Can you try [this](https://github.com/pypa/pipenv/issues/913#issuecomment-336757821)?

lk-geimfari:
So, I have tried to fix the error which raises on Windows. I have no ******* idea why this does not work.

Here is log file:

[log.txt](https://github.com/lk-geimfari/mimesis/files/1610327/log.txt)

lk-geimfari:
I suggest using [this](https://github.com/pypa/pipenv/blob/65c7b2a79559706df54d8950cb273e2c44f492bf/appveyor.yml) as an example which works fine.

Pipfile
```diff
@@ -0,0 +1,27 @@
...
+pytest-isort = "*"
+pytest-cov = "*"
+"pytest-flake8" = "*"
+pytest-mock = "*"
+mypy = "*"
+"flake8-quotes" = "*"
+"flake8-commas" = "*"
+"flake8-builtins" = "*"
+"flake8" = "*"
+tox = "*"
```
lk-geimfari: We need to remove `tox` form dependencies, as we discussed in #351.

lk-geimfari:
@duckyou Tell me, please, what is your operating system? Is it MS Windows? Can you look at this problem, if your OS is Windows? Thanks!

duckyou:
@lk-geimfari Here is what i have - [History](https://ci.appveyor.com/project/duckyou/mimesis/build/1.0.15/job/9mq0x5u4edylaqnu)
And, omfg, [finnaly](https://ci.appveyor.com/project/duckyou/mimesis/build/1.0.17)...

lk-geimfari:
@duckyou I ran into the same problem and I surrendered on the 15th commit :smile:

Thank you very much for help!

PR #354, opened 2018-01-07
Optimization

lk-geimfari:
- Optimized method `Text.hex_color`
- Optimized method `Address.coordinates`
- Optimized method `Internet.ip_v6`


mimesis/providers/text.py
```diff
@@ -134,11 +134,14 @@ def hex_color(self) -> str:
         :return: Hex color code.
 
         :Example:
-            #D8346B
+            #d8346b
         """
-        letters = '0123456789ABCDEF'
-        color_code = '#' + ''.join(self.random.sample(letters, 6))
-        return color_code
+        return '#{:x}'.format(
```
lk-geimfari: Seems like there is an issue, because sometimes the length of the returned string is 6, but must be 7.

PR #359, opened 2018-01-09
Added Pipfile.lock

lk-geimfari:


lk-geimfari:
So, seems like `AppVeyor` don't want to work with `Pipfile.lock`, but it's really important because sometimes the new version of packages can be without backward-compatibility with the early version. Also, version of some packages does нот clearly fixed in `Pipfile`.

lk-geimfari:
I don't think that the problem in `Pipenv`, because it works fine on `Linux`. Possible, that our configuration ([appveyor.yml](https://github.com/lk-geimfari/mimesis/blob/master/appveyor.yml)) is not completely correct.

duckyou:
@lk-geimfari I installed windows 10 on virtualbox and reproduce the error from appveyor. Then i tried create lock file from win and found `colorama` dependency.  :yum:

lk-geimfari:
@duckyou Great job! Thank you very much!

PR #360, opened 2018-01-10
Fixed #351

### LINKED ISSUE #351: Change current development workflow
I have not seen the project in a while and have not contributed to it. I was confused about different tools and workflow this project uses:

1. `setup.py` commands
2. `Makefile`
3. `tox`
4. custom commands in CI

What do I suggest:
1. Stick to one thing only. It might be whatever we choose, but this tool should be the only possible way to do things
2. Update `Contributing.md`, since it does not contain any helpful information. What should be there? A step by step instruction how to do things. How to install dependencies? How to test your code? How to upload it to `pypi`?

lk-geimfari:
As we discussed with @sobolevn, in #351.


lk-geimfari:
@sobolevn Your comments, please. Maybe must be add something else or must be removed something. I'm open to criticism.

lk-geimfari:
@sobolevn @duckyou  Your criticism, please.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+We use `Pipenv` for managing development packages:
+```
+~ pip install pipenv
+~ pipenv install --dev
+```
 
-- Check your spelling and grammar.
+### Code Style
+1. Every contributor must follow the [PEP8](https://www.python.org/dev/peps/pep-0008/) code style.
+2. Always use single-quoted strings, unless a single-quote occurs within the string.
```
sobolevn: This is not very important. Also, this will be checked by linter.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+3. Always run tests, because tests check code style too (`pytest-flake8`).
```
sobolevn: How? Is it possible to check style only? 
Could you please update this part?

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
 
-- Check code with [pycodestyle](https://github.com/PyCQA/pycodestyle) or with any other similar tool.
+### Annotating
+We are using optional static typing (`mypy`) and this means that every contributor should
+annotate his functions and methods (not variables, because Python 3.5 does not support annotation of variables).
```
sobolevn: I suggest to rephrase this part as: "We use optional static typing (`mypy`). Every function and method should be annotated."

Also link to the `mypy` docs would be awesome.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
 
-- Run the tests after each commit.
+```python
+def plus(a: int, b: int) -> int:
+    """Get sum of a and b.
```
sobolevn: This line violates `pydocstyle`. Do we support it? https://gitlab.com/pycqa/flake8-docstrings
If not - that what we could possibly implement sometime soon.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+    :param a: First number.
+    :param b: Second number.
+    :return: Sum of a and b.
+    """
+    return a + b
+```
 
-- If you add any functionality, then you should add tests for it.
+### Documenting
+Always add docstrings for your functions, methods, and classes:
```
sobolevn: This should be checked by linter. Not by a developer.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+    :return: Pi
+    """
+    return 3.14
+```
+
+It's a good idea to add a comment to code line where what is happening is not clear from the context,
+but you shouldn't add comments to pretty obvious things.
+
+### Documentation
+If you have added a new class, module or function than ones should be added to the `docs/api.rst`.
```
sobolevn: Do we measure doc-coverage?

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+
+def pi() -> float:
+    """Get Pi with two digits after the decimal point.
+
+    :return: Pi
+    """
+    return 3.14
+```
+
+It's a good idea to add a comment to code line where what is happening is not clear from the context,
```
sobolevn: Maybe a rephrase? Like: "Comment only things that are not obvious: hacks, optimizations, complex algorithms. Obvious code does not require any additional comments."

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+
+### Testing
+You should write the test which shows that the bug was fixed or that the feature works as expected,
+run test before you commit your changes to the branch and create PR.
+
+To run tests, simply:
+```
+make test
+# or
+~ py.test --benchmark-skip --color=yes ./
```
sobolevn: There should be one way to do it. Just `make test` is fine.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+```
+
+Check out logs of Travis CI or AppVeyor if tests were failed on creating PR, there you can find useful information.
+
+### Performance
+It's good idea to run benchmark test, when you add your feature:
```
sobolevn: Should we do it anyway? Maybe a CI job? 
Sometimes performance degradation is not obvious from the code.

CONTRIBUTING.md
```diff
@@ -1,21 +1,132 @@
...
+
+```
+~ make benchmarks
+# or
+~ py.test -rf --benchmark-only --benchmark-sort=MEAN ./benchmarks
```
sobolevn: The same here. `make benchmarks`.

CONTRIBUTING.md
lk-geimfari: @sobolevn Not yet. Is there a library for this?

CONTRIBUTING.md
lk-geimfari: @sobolevn Should we? What do you suggest?

CONTRIBUTING.md
lk-geimfari: @sobolevn At this moment our benchmarks coverage only the default locale. I don't know should we or not.

CONTRIBUTING.md
sobolevn: https://pypi.python.org/pypi/docstring-coverage/0.3.4 I have not tried it yet.

lk-geimfari: I have created different issues for doc-coverage and docstrings, #364.

sobolevn: `docstring-coverage` seems outdated.

lk-geimfari: @sobolevn I'll try to find something else. At this moment I have applied all requested changes in this branch. I'll update some section of contributing.md when I finish with #361, #364.

PR #365, opened 2018-01-13
[WIP] Fixed #361

### LINKED ISSUE #361: PEP 257
We should integrate some tool, which will check `__doc__` automatically. Your suggestions, gentlemen.

lk-geimfari:


lk-geimfari:
@codecov What is wrong with you?

Issue #335, opened 2017-12-27
Add tests for seeded providers


lk-geimfari:
As we discussed with @duckyou. We should add tests for seeded providers.

duckyou:
@lk-geimfari use seed tests design like i say here https://github.com/mimesis-lab/discussion/issues/1#issuecomment-353059935?
or is there a more right solution?

lk-geimfari:
@duckyou Actually I have not a better solution right now. @sobolevn maybe you can recommend something? What do you think about solution above?

duckyou:
@lk-geimfari, @sobolevn Example of seed tests: https://github.com/duckyou/mimesis/commit/c59e5d8db6fc9ecd3ba3dd53f5ae4ea1a2636d07
With your approval, I will continue to write tests 😉

lk-geimfari:
@duckyou Looks pretty clear to me. Anyway, I don't see another clearer way to tests seed than this. I'm for, if @sobolevn does not offer something else.

lk-geimfari:
@duckyou So, you can continue to write tests for all providers. Thanks!

PR #366, opened 2018-01-13
Adding seed tests #335

duckyou:
1. Grouped all tests using classes.
2. Added tests for seeded data providers.


tests/test_data/test_address.py
```diff
@@ -10,151 +10,220 @@
...
+            assert a1.street_suffix() == a2.street_suffix()
+
+    def test_address(self, _addresses):
+        a1, a2 = _addresses
+        for _ in range(self.TIMES):
+            assert a1.address() == a2.address()
+
+    def test_state(self, _addresses):
+        a1, a2 = _addresses
+        for _ in range(self.TIMES):
```
lk-geimfari: Are there no other much elegant way to repeat test `n` time? Something like:
```
@pytest.mark.repeat(self.TIMES)
def test_state(self, _addresses):
    a1, a2 = _addresses
    assert a1.state() == a2.state()
    assert a1.state(abbr=True) == a2.state(abbr=True)
```

sobolevn: Why do even need to repeat a test?

mimesis/providers/base.py
```diff
@@ -9,7 +9,7 @@ class BaseProvider(object):
     """This is a base class for all providers.
     """
 
-    def __init__(self, seed: Optional[int] = None) -> None:
+    def __init__(self, seed: Union[int, str, bytes, bytearray] = None) -> None:
```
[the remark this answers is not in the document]
sobolevn: It is better to use a custom type here:

```python
Seed = Union[int, str, bytes, bytearray]  # or whatever it is
```

tests/test_data/test_address.py
lk-geimfari: I think to ensure that result is same for each iteration.

mimesis/providers/internet.py
```diff
@@ -119,9 +119,8 @@ def emoji(self) -> str:
         """
         return self.random.choice(EMOJI)
 
-    @staticmethod
-    def image_placeholder(width: Size = 400,
-                          height: Size = 300) -> str:
+    # TODO: https://en.wikipedia.org/wiki/List_of_common_resolutions
```
sobolevn: You can specify `todo`s in code: https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html

lk-geimfari:
```rst
.. todo:: https://en.wikipedia.org/wiki/List_of_common_resolutions
```

tests/test_data/test_address.py
duckyou: @lk-geimfari If I understand `pytest` correctly, when we writing tests with decorator, providers initialization at runtime will be at every iteration.

lk-geimfari: @duckyou You're right. My idea was not really correct or rather really incorrect.

lk-geimfari:
@duckyou Can we merge it? Is it done?

duckyou:
@lk-geimfari It seems like yes 😄 
@lk-geimfari and @sobolevn Also big thanks for your reviews, guys!  :wink:

lk-geimfari:
@duckyou Thank you for great work!

PR #374, opened 2018-01-18
Fixed #353

### LINKED ISSUE #353: Editorconfig is incorrect
For some reason `editorconfig` is not applied to many files in the project. `.travis.yml`, `appveyor.yml` to name a few. These files uses spaces (that's correct), but they should use tabs as it is stated in the `editorconfig` (that's incorrect).

Only `Makefile` should use tabs, other files are good with spaces.
Also, related to the #351 as we should document how to integrated `editorconfig` into the workflow.

lk-geimfari:


lk-geimfari:
@sobolevn I have not applied this to files using an auto-formatting tool (PyCharm, `ctrl + alt + shift`). I'll do it after your review.

.editorconfig
```diff
@@ -4,11 +4,22 @@ root = true
...
 end_of_line = lf
+indent_style = space
 insert_final_newline = true
 
 [*.py]
-indent_style = space
 indent_size = 4
+
+[*{json,yml}]
+indent_style = space
```
lk-geimfari: I'm not sure about this because file indented with tabs is smaller than a file indented with spaces. The size of the file is important when it comes to JSON.

Makefile
```diff
@@ -64,6 +64,11 @@ publish:
 	python3 setup.py sdist && twine upload dist/*
 
 
+.PHONY:	style-check
```
sobolevn: Extra spaces

Makefile
```diff
@@ -49,7 +49,7 @@ test:
 	py.test --benchmark-skip --color=yes ./
 
 
-.PHONY: type-check
+.PHONY:type-check
```
sobolevn: missing space

PR #378, opened 2018-01-20
Added explicit definition of provider

lk-geimfari:
Some data providers have methods with the same name (`Personal.title` and `Text.title`) and previous version of `schema.Field` always returns a result of the method `title()` of provider `Personal()`:
```python
>>> from mimesis.schema import Field
>>> _ = Field('en')
>>> _('title')
'Mr.'
```
Now we can explicitly define that the method belongs to data-provider ``cls``:

```python
>>> _('title', cls='text')
'She spent her earliest years reading classic literature.'
>>> _('title', cls='personal')
'Mrs.'
```

---

In addition, an explicit definition significantly increases the speed of work:

**The result on defined `cls`**:
```python
_ = Field('en')
description = (
    lambda: {
        'id': _('cryptographic.uuid'),
        'name': _('text.word'),
        'version': _('development.version', pre_release=True),
        'timestamp': _('datetime.timestamp', posix=False),
        'owner': {
            'email': _('personal.email', key=str.lower, domains=['@example.com']),
            'token': _('cryptographic.token'),
            'creator': _('personal.full_name', gender=Gender.FEMALE),
        },
        'title': _('text.title')
    }
)
schema = Schema(schema=description)
cProfile.run('schema.create(iterations=1000)')
```
```
         269711 function calls (269584 primitive calls) in 0.549 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1000    0.018    0.000    0.547    0.001 .test.py:10(<lambda>)
        # ... shortened result.
```

**The result on undefined `cls`:**
```python
_ = Field('en')
description2 = (
    lambda: {
        'id': _('uuid'),
        'name': _('word'),
        'version': _('version', pre_release=True),
        'timestamp': _('timestamp', posix=False),
        'owner': {
            'email': _('email', key=str.lower, domains=['@example.com']),
            'token': _('token'),
            'creator': _('full_name', gender=Gender.FEMALE),
        },
        'title': _('title')
    }
)
schema2 = Schema(schema=description2)
cProfile.run('schema2.create(iterations=1000)')
```
```
         1283777 function calls (1283650 primitive calls) in 4.868 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1000    0.031    0.000    4.865    0.005 .test.py:25(<lambda>)
        # ... shortened result.
```
----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Version 1.0.6
+
+**Added**:
+- Explicit definition of data-provider for `schema.Field`
+
+---
+
 ## Version 1.0.5
 
 **Added**:
--- a/mimesis/schema.py
+++ b/mimesis/schema.py
@@ -44,8 +44,13 @@ class AbstractField(StrMixin):
         represents name of the any method of any supported data provider
         and the ``**kwargs`` of this method.
 
+        ..note:: Some data providers have methods with the same name and
+        in such cases, you can explicitly define that the method belongs to
+        data-provider ``name='provider.name'``.
+
         You can apply a *key function* to result returned by the method,
-        to do it, just pass a callable object which returns final result.
+        to do it, just pass parameter ``key`` with a callable object which
+        returns final result.
 
         :param name: Name of the method.
         :param key: A key function (or other callable object)
@@ -58,8 +63,13 @@ class AbstractField(StrMixin):
         if name is None:
             raise UndefinedField()
 
-        # TODO: This is a really slow solution. Fix it.
-        for provider in dir(self.gen):
+        providers = []
+        if '.' in name:
+            cls, name = name.split('.', 1)
+            providers.append(cls)
+
+        # TODO: Optimization.
+        for provider in providers or dir(self.gen):
             if hasattr(self.gen, provider):
                 provider = getattr(self.gen, provider)
                 if name in dir(provider):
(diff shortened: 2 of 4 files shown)


lk-geimfari:
I'm open to criticism.

sobolevn:
@lk-geimfari can you please test the version with passing real classes, not strings?
How much faster/slower would that be?

lk-geimfari:
@sobolevn It will not be easy: [#L37](https://github.com/lk-geimfari/mimesis/blob/optimized-schema/mimesis/schema.py#L37), [#L69](https://github.com/lk-geimfari/mimesis/blob/optimized-schema/mimesis/schema.py#L69).

duckyou:
@lk-geimfari imho, with `cls` parameter we lost feature of `Scheme` provider...
Here is example of using generic for retrieve same result:
```python
g = Generic('en')

desc = lambda: {
    'id': g.cryptographic.uuid(),
    'name': g.text.word(),
    'version': g.development.version(pre_release=True),
    'owner': {
        'email': g.personal.email(domains=['@example.com']).lower(),
        'token': g.cryptographic.token(),
        'creator': g.personal.full_name(gender=Gender.FEMALE),
    },
    'title': g.text.title(),
}

print(desc())
```
Maybe we should specify provider in string?
Something like `_('text.title')`... we can skip this for unique-named methods like `_('uuid')`

lk-geimfari:
@duckyou I like it! `text.title` looks nice.

PR #380, opened 2018-01-21
Fixed error with distribution in an equation.

dynamicwebpaige:
image


lk-geimfari:
@dynamicwebpaige Thank you! Unfortunately, our scientific data is really poor and нот always accurate.

Issue #387, opened 2018-01-22
Generate geographic coordinates for country territory only


lk-geimfari:
When locale is specified it would be awesome to generate coordinates (`Address.coordinates`) only on the territory of the specified country.

**UPD:**

We can use [this](https://gist.github.com/lk-geimfari/0e5638302b00f299b39066223906e7d0) gist.

lk-geimfari:
@duckyou Any ideas?

pskalyber:
I we need it, we can use Google dataset here: https://developers.google.com/public-data/docs/canonical/countries_csv

lk-geimfari:
The second question is should we add this feature only for supported locales or optionally, i.e like that:

```python
coordinates(country_code=Coordinates.RU)
```

lk-geimfari:
@pskalyber Do you know how to use this ranges? I really don't know how to do it.

For example, we have coordinates of USA:
```
"US": {"lt": 37.09024, "lg": -95.712891}
```

How can we generate random `latitude` for USA? What's range should we use?

PR #389, opened 2018-01-23
[WIP] Implemented #387

lk-geimfari:
How it works:

When we define locale, methods just returns approximate coordinates for current locale (country):
```python
>>> ad = Address('de') # It will return coordinates for Germany
>>> ad.coordinates()
{'latitude': 47.523907, 'longitude': 12.963119}
```
screenshot from 2018-01-23 14-36-26

But I don't know how to be with passing an optional range of coordinates. For example, we have created an instance for de, but want to generate random coordinates.

Should we use do it like this:
```python
>>> ad.coordinates(country_code='any')
```

Or, maybe like this:

```python
>>> ad.coordinates(country_code=...)
```


lk-geimfari:
@sobolevn @duckyou Your opinion, please.

duckyou:
@lk-geimfari Coordinates is **locale pattern** independent data, but sometimes it should be depend on **country locale**... hmm... Maybe migrate this method to international provider?

lk-geimfari:
@duckyou I have added ones to JSON files for easy creating data for a defined locale. Maybe you're right. Can you please suggest your solution, when you'll have free time. All the needed data you can grab forked this branch.a

lk-geimfari:
My solution looks not good.

PR #390, opened 2018-01-23
Add some data generators for RussiaSpecProvider

axce1:
Add OGRN and BIC data generators for RussiaSpecProvider #318


lk-geimfari:
@axce1 Great! Thank you for your contribute!

PR #391, opened 2018-01-23
Updated tests

lk-geimfari:
What was be done:

- [x] Refactored some tests.
- [x] Simplified testing of seeded providers.

Why:

It was too hard to maintain the previous tests.

### DIFF
--- a/tests/test_providers/test_address.py
+++ b/tests/test_providers/test_address.py
@@ -9,16 +9,17 @@ from mimesis.data import CALLING_CODES, CONTINENT_CODES, COUNTRIES_ISO
 from mimesis.enums import CountryCode
 from mimesis.exceptions import NonEnumerableError
 
-from . import _patterns as p
+from ._patterns import POSTAL_CODE_REGEX, STR_REGEX
 
 
 class TestAddress(object):
+
     @pytest.fixture
     def _address(self):
         return Address()
 
     def test_str(self, address):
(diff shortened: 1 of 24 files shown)


tests/test_providers/test_address.py
```diff
@@ -149,80 +149,53 @@ def test_calling_code(self, _address):
...
-    def test_city(self, _addresses):
-        a1, a2 = _addresses
-        for _ in range(self.TIMES):
-            assert a1.city() == a2.city()
-
-    def test_continent(self, _addresses):
-        a1, a2 = _addresses
-        for _ in range(self.TIMES):
-            assert a1.continent() == a2.continent()
-            assert a1.continent(code=True) == a2.continent(code=True)
```
duckyou: @lk-geimfari Check seeded provider methods with arguments also neccesary. Because there may be specific conditions when we pass arguments to method. When i wrote seed tests in first time there were many such situations.

lk-geimfari: @duckyou I'll revert changes related to such cases.

lk-geimfari: @duckyou I have revert changes. Also, we must think about doing library easily extensible. We must get rid of the hard-coded parts of the code and I hope to your help.

lk-geimfari: I have created #392

PR #412, opened 2018-03-03
[tests] Fix a broken test related to version control systems

aprasanna:
Add missing 'Mercurial' entry

### DIFF
--- a/tests/test_providers/test_development.py
+++ b/tests/test_providers/test_development.py
@@ -22,7 +22,7 @@ class TestDevelopment(object):
         assert result in data.LICENSES
 
     def test_version_control_system(self, dev):
-        vcs = ['Git', 'Subversion', 'Mercurial']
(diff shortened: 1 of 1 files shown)


tests/test_providers/test_development.py
```diff
@@ -22,7 +22,7 @@ def test_license(self, dev):
         assert result in data.LICENSES
 
     def test_version_control_system(self, dev):
-        vcs = ['Git', 'Subversion']
+        vcs = ['Git', 'Subversion', 'Mercurial']
```
lk-geimfari: I think that is better to use `data.VERSION_CONTROL_SYSTEMS`

lk-geimfari: @aprasanna Look at review, please.

aprasanna:
Agreed. That looks more correct and clean. Thanks!

PR #431, opened 2018-04-02
Fixes setup.cfg and Pipfile

sobolevn:
Our `setup.cfg` had several issues. I have fixed them.
You can check our latest version: https://github.com/wemake-services/wemake-django-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/setup.cfg

I have also fixed `Pipfile`. And style issues in `setup.py`.


sobolevn:
Broken due to #432

lk-geimfari:
@sobolevn I have tried to fix an issue with mypy, but seems I have done something wrong. If you can fix it, then do it, please. Thank you!

sobolevn:
Ok, then I am merging this. And then fixing `mypy` issues.

PR #464, opened 2018-06-16
change Unsplash image url with more relevance results

RaidoS:
Previous url returns not relevance image.
before:
https://source.unsplash.com/category/sea/1900x1080
now:
https://source.unsplash.com/1900x1080/?sea

in addition, category (or search keyword) can be coma separated list, like so https://source.unsplash.com/1900x1080/?sea,sun

Unsplash reference https://source.unsplash.com/


lk-geimfari:
@RaidoS Looks good. Thank you!

PR #481, opened 2018-07-13
Added generate KPP (issue #318)

Vlangf:
Added generate KPP


lk-geimfari:
@Vlangf Great job! Thank you for your contribution!

Issue #485, opened 2018-07-21
Datetime timezone support


sobolevn:
Currently I am unable to set timezone for cases like this:

```python
processed_at = mimesis('datetime')
```

But I would like to do something like:

```python
# To select random tz:
processed_at = mimesis('datetime', timezone=True)  # True should be a default, here I am just explicitly calling it

processed_at = mimesis('datetime', timezone=some_specific_tz)
```

I am not sure about the dependecies. Timezone support out-of-the-box is limited, we almost always use something like `pytz`.

So, maybe in this case we will just implement the second case.
Leaving timezone support optional. And only enabling passed in timezone.

northernwulf:
Hi.
As I understood it will be a another property zone: Optional[str] = None in constructor of BaseDataProvider. Also one more question: should it be a separate method for getting datetime with timezone or we need to add it to existent?

lk-geimfari:
@northernwulf Hi! We already have method [datetime()](https://github.com/lk-geimfari/mimesis/blob/master/mimesis/providers/date.py#L179) and all we need is just add support of parameter timezone for it. 

@sobolevn Correct me if I'm wrong.

sobolevn:
@lk-geimfari it seems correct to me.

northernwulf:
@lk-geimfari got it. Thanks

e-vandenberg:
Hi! Just wondering if this enhancement has been added yet.

If not I'd like to give it a shot.

lk-geimfari:
@e-vandenberg Hi! It still did not implement you can take it and implement!

lk-geimfari:
@sobolevn Please, share with your vision of this feature with @e-vandenberg

sobolevn:
@e-vandenberg yeap, I had something similar in mind.

lk-geimfari:
Implemented in #539

PR #494, opened 2018-08-25
Implemented timezone support, #485

e-vandenberg:
Added feature from issue #485 to allow custom timezones with datetime.

Below is an example of usage:

```python
>>> d = Datetime()
>>> d.datetime(timezone="Europe/Paris")
datetime.datetime(2003, 12, 18, 4, 53, 48,  tzinfo=<DstTzInfo 'Europe/Paris' CEST+2:00:00 DST>)
```

This enhancement requires pytz.

### DIFF
--- a/mimesis/providers/date.py
+++ b/mimesis/providers/date.py
@@ -2,12 +2,12 @@
 
 from calendar import monthrange, timegm
 from datetime import date, datetime, time
-import pytz
 
 from mimesis.data import GMT_OFFSETS, ROMAN_NUMS, TIMEZONES
 from mimesis.providers.base import BaseDataProvider
 from mimesis.typing import DateTime, Timestamp
 from mimesis.utils import pull
+from mimesis.compat import pytz
 
 __all__ = ['Datetime']
 
@@ -201,8 +201,11 @@ class Datetime(BaseDataProvider):
             return dt.strftime('%B, %d %Y')
 
         if timezone != '':
-            timezone = pytz.timezone(timezone)
-            dt = timezone.localize(dt)
+            if not pytz:
+                raise ImportError('Timezones are supported only with pytz')
+            else:
+                timezone = pytz.timezone(timezone)
+                dt = timezone.localize(dt)
 
         return dt
 
--- a/mimesis/providers/date.py
+++ b/mimesis/providers/date.py
@@ -200,7 +200,7 @@ class Datetime(BaseDataProvider):
         if humanized:
             return dt.strftime('%B, %d %Y')
 
-        if timezone != '' and not humanized:
+        if timezone != '':
             if not pytz:
                 raise ImportError('Timezones are supported only with pytz')
             else:
--- a/mimesis/providers/date.py
+++ b/mimesis/providers/date.py
@@ -2,6 +2,7 @@
 
 from calendar import monthrange, timegm
 from datetime import date, datetime, time
+import pytz
 
 from mimesis.data import GMT_OFFSETS, ROMAN_NUMS, TIMEZONES
 from mimesis.providers.base import BaseDataProvider
@@ -176,10 +177,11 @@ class Datetime(BaseDataProvider):
         """
         return self.random.choice(GMT_OFFSETS)
 
-    def datetime(self, humanized: bool = False, **kwargs) -> DateTime:
+    def datetime(self, humanized: bool = False, timezone: str = '', **kwargs) -> DateTime:
         """Generate random datetime.
 
         :param humanized: Readable representation.
+        :param timezone: Custom timezone.
         :param kwargs: Keyword arguments (start, end).
         :return: Datetime.
         :rtype: datetime.datetime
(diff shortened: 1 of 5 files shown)


lk-geimfari:
We can't add dependencies only for this little feature, os i think that must be optional feature, I suggest something like this:

```python

    def datetime(self, humanized: bool = False, timezone: str = '', **kwargs) -> DateTime:
        """Generate random datetime.
        :param humanized: Readable representation.
        :param timezone: Custom timezone.
        :param kwargs: Keyword arguments (start, end).
        :return: Datetime.
        :rtype: datetime.datetime
        :Example:
            March, 24 2002
        """
        fmt = '%Y-%m-%d %H:%M:%S'
        dt_str = '{date} {time}'.format(
            date=self.date(fmt='%Y-%m-%d', **kwargs),
            time=self.time(),
        )

        dt = datetime.strptime(dt_str, fmt)

        if humanized:
            return dt.strftime('%B, %d %Y')

        if timezone != '':
            try:
                import pytz
                timezone = pytz.timezone(timezone)
                return  timezone.localize(dt)
            except ImportError:
              raise ImportError('Timezones supported only with pytz')
        return dt
```

@sobolevn What do you think?

sobolevn:
Agreed about not adding a direct dependency.

I guess we can create some kind of `compat.py` file with the following contents:

```python
try:
    import pytz
except ImportError:
   pytz = None
```

And then use it like so:

```python
from mimesis.compat import pytz

def datetime(self, humanized: bool = False, timezone: str = '', **kwargs) -> DateTime:
       ...

        if timezone:
            if not pytz:
                raise ImportError('Timezones are supported only with pytz')
           ...
        return dt
```

lk-geimfari:
@sobolevn Totally agree. @e-vandenberg Please, look at the comment above.

mimesis/providers/date.py
```diff
@@ -176,10 +177,11 @@ def gmt_offset(self) -> str:
         """
         return self.random.choice(GMT_OFFSETS)
 
-    def datetime(self, humanized: bool = False, **kwargs) -> DateTime:
+    def datetime(self, humanized: bool = False, timezone: str = '', **kwargs) -> DateTime:
         """Generate random datetime.
 
         :param humanized: Readable representation.
+        :param timezone: Custom timezone.
```
sobolevn: We need to add a note about `pytz` here. We should also mention that installing `pytz` is required to use it.

lk-geimfari:
@e-vandenberg Also, please don't forget about tests. You can add `pytz` to dev dependencies and test this feature.

e-vandenberg:
@lk-geimfari @sobolevn Thank you both very much for the feedback and support. Is there anything else I should add to this PR?

mimesis/providers/date.py
```diff
@@ -198,6 +202,13 @@ def datetime(self, humanized: bool = False, **kwargs) -> DateTime:
         if humanized:
             return dt.strftime('%B, %d %Y')
 
+        if timezone != '':
+            if not pytz:
+                raise ImportError('Timezones are supported only with pytz')
+            else:
```
sobolevn: We do not need `else` here.

tests/test_providers/test_datetime.py
```diff
@@ -113,6 +114,13 @@ def test_datetime(self, _datetime, start, end, humanized, _type):
             year = int(dt.split(' ')[2])
             assert year == 2018
 
+        if humanized == True:
```
sobolevn: Unneeded condition.

sobolevn:
Please, also, pay attention to the travis build. Some checks are failing.

tests/test_providers/test_datetime.py
e-vandenberg: If humanized is passed as `True` there will be no `tzinfo` attribute and the proceeding lines will result in an error. This `if ... pass` makes sure `tzinfo` isn't checked when the datetime is humanized. 

Maybe I am missing something, but I can't seem to find a better way to do this. Did you have anything in mind?

e-vandenberg:
I have fixed the formatting issues, and most of the bugs, however, there are two remaining errors (as seen in both the Travis CI and AppVeyor tests) that I have not been able to resolve. 

- The first is in `test_datetime.py` and shows no error message, however it comes up as a failure. I am not sure what to make of this. Below is he test result.
```python
def test_datetime(self, _datetime, start, end, humanized, timezone, _type):
        dt = _datetime.datetime(start=start, end=end,
>                                               humanized=humanized, timezone=timezone)
```
- The second one is an `ImportError` that is called in the program. I believe to fix this, `pytz` needs to be added properly to the `Pipfile` and `Pipfile.lock`, however I am having trouble adding to the `Pipfile.lock`

Any help with these small issues would be greatly appreciated.

sobolevn:
@e-vandenberg try to `pipenv install --dev pytz`. This will add `pytz` to both `Pipefile` and `Pipfile.lock`. 

I have also opened a new issue about our documentation missing "Adding dependencies" section: https://github.com/lk-geimfari/mimesis/issues/498

e-vandenberg:
Travis CI checks passed, some AppVeyor checks passed. Do all checks need to pass?

lk-geimfari:
@e-vandenberg Can you please recreate PR?

lk-geimfari:
I have add it by myself in #539

PR #499, opened 2018-09-03
Fixed issue with names on downloading images

conformist-mw:
I have fixed function `download_image`. Also fixed if `save_path` without trailing slash.

### DIFF
--- a/mimesis/utils.py
+++ b/mimesis/utils.py
@@ -121,10 +121,9 @@ def download_image(url: str = '', save_path: str = '',
             image_name = '{}.jpg'.format(uuid4())
         else:
             image_name = '{}.{}'.format(uuid4(), splitted_name[-1])
-        if save_path and not save_path.endswith('/'):
-            save_path = '{}/'.format(save_path)
-        request.urlretrieve(url, save_path + image_name)
-        return image_name
+        full_image_path = path.join(save_path, image_name)
+        request.urlretrieve(url, full_image_path)
+        return full_image_path
     return None
 
 
--- a/mimesis/utils.py
+++ b/mimesis/utils.py
@@ -7,6 +7,7 @@ import ssl
 from os import path
 from typing import Mapping, Optional, Union
 from urllib import request
+from uuid import uuid4
 
 from mimesis import config
 from mimesis.exceptions import UnsupportedLocale
@@ -114,6 +115,14 @@ def download_image(url: str = '', save_path: str = '',
 
     if url:
         image_name = url.rsplit('/')[-1]
+
+        splitted_name = image_name.rsplit('.')
+        if len(splitted_name) < 2:
+            image_name = '{}.jpg'.format(uuid4())
+        else:
+            image_name = '{}.{}'.format(uuid4(), splitted_name[-1])
+        if save_path and not save_path.endswith('/'):
+            save_path = '{}/'.format(save_path)
         request.urlretrieve(url, save_path + image_name)
         return image_name
     return None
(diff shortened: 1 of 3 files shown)


mimesis/utils.py
```diff
@@ -114,6 +115,13 @@ def download_image(url: str = '', save_path: str = '',
 
     if url:
         image_name = url.rsplit('/')[-1]
+
+        # check if image name has two parts: name and ext to save it properly
```
lk-geimfari: I guess that it's better to always generate a random name for all images.

lk-geimfari: @conformist-mw What do you think? Can we remove comment and add random names for every downloaded image?

lk-geimfari:
@conformist-mw So, I have only one comment and if you can fix it and update tests for it then we can merge it.

P.S Also, don't forget to add yourself to our list of contributors it's important.

conformist-mw:
Added generating filename always and fix test.

lk-geimfari:
@conformist-mw Any idea why AppVeyor tests failed?

conformist-mw:
Error is: `Build cloud 'lw-host-14-e' does not support 'Visual Studio 2015' image.` and I have no idea what is mean.

lk-geimfari:
@conformist-mw How about return not image name, but a path to the downloaded image? I think it will be more useful. What do you think?

lk-geimfari:
@conformist-mw Don't forget about docstings: `Path to downloaded image`, instead `Image name`.

mimesis/utils.py
```diff
@@ -106,16 +107,23 @@ def download_image(url: str = '', save_path: str = '',
     :param url: URL to image.
     :param save_path: Saving path.
     :param unverified_ctx: Create unverified context.
-    :return: Image name.
+    :return: Path to download image.
```
lk-geimfari: A path to the downloaded image.

lk-geimfari:
@conformist-mw So, last comment and we can merge this PR. Thank you for your patience and interest.

mimesis/utils.py
conformist-mw: Sorry :)

lk-geimfari:
@conformist-mw Also seems like we don't cover all cases of images. I mean we should test images with/without a name too, to be sure that everything is okay.

lk-geimfari:
@conformist-mw I'm sorry about that. I hope that my comments do not irritate you.

mimesis/utils.py
```diff
@@ -106,16 +107,23 @@ def download_image(url: str = '', save_path: str = '',
...
         ssl._create_default_https_context = ssl._create_unverified_context
 
     if url:
         image_name = url.rsplit('/')[-1]
-        request.urlretrieve(url, save_path + image_name)
-        return image_name
+
+        splitted_name = image_name.rsplit('.')
+        if len(splitted_name) < 2:
+            image_name = '{}.jpg'.format(uuid4())
```
lk-geimfari: We should test this case

mimesis/utils.py
```diff
@@ -106,16 +107,23 @@ def download_image(url: str = '', save_path: str = '',
...
+        else:
+            image_name = '{}.{}'.format(uuid4(), splitted_name[-1])
```
lk-geimfari: And this

mimesis/utils.py
conformist-mw: Added.

mimesis/utils.py
conformist-mw: Already tested.

lk-geimfari:
@conformist-mw Great! Thank you!

PR #516, opened 2018-09-27
Added OSX support

lk-geimfari:
I have added OSX to TravisCI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -189,3 +189,5 @@ fabric.properties
 # mypy
 .mypy_cache/
 .test.py
+
+.pytest_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,11 @@ matrix:
           python: "3.7"
           dist: xenial
 
+        - os: linux
+          sudo: required
+          python: "pypy3.5"
+          dist: xenial
+
         - os: osx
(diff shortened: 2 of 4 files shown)


lk-geimfari:
@sobolevn Any comments?

sobolevn:
@lk-geimfari dropping `3.5` support is a big thing.
What are the reasons for this step?

lk-geimfari:
@sobolevn Full type hints support, including variables.

sobolevn:
@lk-geimfari do we have any kind of variables that can not be annotated for now?
Can `# type: some_type` fix these cases?

lk-geimfari:
@sobolevn We will add note to changelog to make sure that users understand that Python 3.5 is no longer supported.

We can use `# type: some_type`, but this looks a little bit annoying.

Why do I suggest reject of Python 3.5:

1. To add full type hints support.
2. Most distributives uses Python 3.6^ as standard interpreter.
3. We will update a lot of code anyway to make it much clearer*.

** I have planned some big changes and updates and possible in API (of course I will not turn it upside down). Next releases will not contain any breaking changes after final migration to Python 3.6, so I think that we should do it (reject using Python 3.5) now, with other a little breaking changes

Issue #468, opened 2018-06-16
Add pytest-randomly 


sobolevn:
We need to run tests in random order to be sure, that they do not depend on each other.

https://github.com/pytest-dev/pytest-randomly

lk-geimfari:
@sobolevn Seems like this already fixed.

sobolevn:
No, it is not fixed. https://github.com/pytest-dev/pytest-randomly/issues/135

lk-geimfari:
So, we can implement it now: https://github.com/pytest-dev/pytest-randomly/pull/187

lk-geimfari:
@sobolevn Do you know how to add entry point to mimesis? I'm not sure that I completely understand the idea of how `pytest-randomly` works with entry points.

sobolevn:
Nope, sorry, I haven't seen it yet.

tabassco:
If no one is working on this at the moment, I'd like to give it a shot.

tabassco:
Checking this issue vs #469 I see it that this issue is already fulfilled. As described in the opening post, the intent was to run pytest in random order to check for possible dependencies of the tests. 

To make use of the seed functionality in pytest-randomly I modified the makefile so that it now takes an optional parameter SEED. 

```
SEED = unspecified

ifeq ("$(SEED)", "unspecified")
RANDOM_SEED =
else
RANDOM_SEED = --randomly-seed=$(SEED)
endif
```

```
.PHONY: test
test:
	pytest --color=yes $(RANDOM_SEED) ./
	mypy mimesis/ tests/
	make clean
```


SEED can be:
 "last" -> rerunning the tests with the seed from the previous run
  int -> giving specific seed

As I'm mainly a Python developer I'm not that well versed in makefiles and would appreciate your modifications in making it prettier.

Issue #469, opened 2018-06-16
Support pytest-randomly


sobolevn:
It is said, that `pytest-randomly` supports `faker` and `factory-boy`:

```
Randomly shuffles the order of test items. This is done first at the level of modules, then at the level of test classes (if you have them), then at the order of functions. This also works with things like doctests.
Resets random.seed() at the start of every test case and test to a fixed number - this defaults to time.time() from the start of your test run, but you can pass in --randomly-seed to repeat a randomness-induced failure.
If factory boy is installed, its random state is reset at the start of every test. This allows for repeatable use of its random 'fuzzy' features.
If faker is installed, its random state is reset at the start of every test. This is also for repeatable fuzzy data in tests - factory boy uses faker for lots of data.
If numpy is installed, its random state is reset at the start of every test.
```

We need to think about possible integration with this project.

sobolevn:
@lk-geimfari friendly ping. Do you have some time to work on this? https://github.com/pytest-dev/pytest-randomly/issues/135

lk-geimfari:
@sobolevn I'm sorry, but I really can't work on any issue at this weekends. I'll try to work on it in next weekends.

lk-geimfari:
@sobolevn Can you, please, tell me what steps exactly should I do?

sobolevn:
It is better to ask `pytest-randomly` maintainers. I don't remember what is the status of this issue.

lk-geimfari:
@sobolevn I hear ya. Already done! Thanks!

lk-geimfari:
Fixed in #849

MKaras93:
Hello, I have a question - is randomly supported on 4.1.3? I'm working on Python 3.6, so I can't install a newer version. I'm using `mimesis==4.1.3` and `pytest-randomly==3.10.3`, however even with argument `--randomly-seed=1` I'm getting different results from mimesis when I rerun test. Other random tools repeat their output as expected, I only have problem with mimesis.

It's possible that the problem is on my side (there are some custom factories which I haven't written, I'm still figuring it out), but it would definitely help to know if mimesis seed can be controlled by randomly in 4.1.3.

I've tried to figure it out from the commit list, but there are quite a few issues mentioning this problem and I'm not sure if they refer to the same thing.

sobolevn:
I have a PR ready ;)

sobolevn:
Fixed by https://github.com/lk-geimfari/mimesis/pull/1325

PR #521, opened 2018-10-03
Add pytest-randomly plugin (#468

destag:


lk-geimfari:
@sobolevn Can you please explain to @destag what do we actually need and how to do it. Thanks!

destag:
@sobolevn thanks for your review. I updated the lockfile.
I think you're talking about issue #469 which is in fact, much more complex. I can try to work on it if you wish.

PR #524, opened 2018-10-04
Update/verify Polish locale data

pywkm:
(related with #96)

Hi,
I know, at times, commits look drastic (8.5k deletions), but I think now data are more Polish than before :) - for instance some names didn't even sound Polish, some of them can be found in Poland, but if basic purpose of `mimesis` is that data should look as localized, better get rid of those uncertain/controversial ones. :)

I have one question for you - every other locale has "bad" words capitalized, also sometimes those are phrases not words. Why is that? It leads to funny situations where word "Satan" or "Devil" is capitalized but "god" is in lowercase. Wait until some freaks find out your project (similar to ones that wipe out from many repositories words "slave" and "master", because of reasons) ;).
Anyway, I wrote every word lowercased, bad words are words not phrases, to be consistent. But I can change that if you want. ;)


lk-geimfari:
@pywkm Lol, they were copy-pasted from the Internet and it's just a coincidence. We are not the minions of Satan. 

If you can, please update list of bad words to make it much clearer.

Dziękuję dziękuję!

lk-geimfari:
@pywkm If that's all let me know and I'll merge this great PR!

lk-geimfari:
@pywkm I almost forgot! Please, add yourself to list of our [contributors](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst) it's important.

pywkm:
Can you explain why the coverage test failed in this PR after rebasing on newest master? :( On the file that I didn't touch. :) Can I do something about this?

Issue #523, opened 2018-10-04
Bug in [doctest] mimesis.utils.pull


Valerievich:
There is a bug in [doctest] mimesis.utils.pull:

```>>> en['day']['abbr'][0]
Expected:
    'Mon.'
Got:
    'Fri.'
```

Valerievich:
I'm going to deal with it

lk-geimfari:
@Valerievich Can you please add more information about how to reproduce this bug?

Valerievich:
Sure! Just run `pytest --doctest-modules` and you'll see `DocTestFailure`

Valerievich:
Can't start fixing this. My [PR](https://github.com/lk-geimfari/mimesis/pull/522) hasn't approved yet.

lk-geimfari:
@Valerievich merged! You can start fixing this.

Valerievich:
@lk-geimfari Since we've merged PR we could close this related issue.

PR #528, opened 2018-10-09
Bugfix in utils.py doctests provided DocTestFailure

Valerievich:
Bugfix this [issue](https://github.com/lk-geimfari/mimesis/issues/523).


Valerievich:
@lk-geimfari  Should I check something more? I think it's save to approve this. There is only one small change.

PR #529, opened 2018-10-09
Fix pipenv with pip==18.1 bug

destag:
There is [bug](https://github.com/pypa/pipenv/issues/2924) with pipenv used with pip==18.1.
Because of that all Travis builds for python3.7 are failing.
I added temporary workaround.


lk-geimfari:
@destag Thank you very much!

Issue #507, opened 2018-09-17
Add support of pre-commit hooks


lk-geimfari:
Can we add support of [pre-cmmi](https://pre-commit.com/)t hooks?

GustavoKatel:
I could add an initial support at #531

PR #532, opened 2018-10-11
Add pre-commit support: flake8 and mypy

GustavoKatel:
Hi! This adds support for pre-commit (#507)

I took the liberty to also add make targets to install the hooks

Please let me know if this is reasonable to everyone or if I should add more hooks to the config.


lk-geimfari:
@GustavoKatel Hi! Can you please add information about how to use it this to `CONTRIBUTING.rst` file? It can help future contributors.

PR #539, opened 2018-10-16
[WIP] v2.2.0

lk-geimfari:
## Checklist

- [x] I have read [contributing guidelines](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTING.rst).
- [x] I'm sure that the cat did not jump to the keyboard and did not unrelated changes in this pull request
- [x] I have added myself to the `CONTRIBUTORS.md`
- [x] I have added my changes to the `CHANGELOG.md`


sobolevn:
Please, do not forget about doctests

lk-geimfari:
@sobolevn Sure! Work still in progress.

lk-geimfari:
@sobolevn How about following format of doctests:

```python
    def street_number(self, maximum: int = 1400) -> str:
        """Generate a random street number.

        :param maximum: Maximum value.
        :return: Street number.

        :Example:
        >>> address = Address()
        >>> number = address.street_number(128)
        >>> isinstance(number, str)
        True
        >>> int(number) <= 128
        True
        """
        return str(self.random.randint(1, maximum))
```

Should i add `:Example:`?

lk-geimfari:
@sobolevn Can you, please, review module `address.py` where added doctests?

sobolevn:
@lk-geimfari LGTM!

sobolevn:
That's not for testing. That's done to be sure that examples are still valid.
We can change something and then miss to fix it in the docs.

These tests ensure that we won't forget about it.

lk-geimfari:
@sobolevn So, then that's a good idea! Should we show doctests in documentations?

sobolevn:
@lk-geimfari yes, we surely have to show doctests in the docs.

mimesis/providers/business.py
```diff
@@ -25,7 +25,11 @@ def company(self) -> str:
         :return: Company name.
 
         :Example:
-            Gamma Systems.
+
+        >>> business = Business()
+        >>> company = business.company()
+        >>> company in business._data['company']['name']
```
sobolevn: We need to see a real value here. Like `'Microsoft'`. We can use `random.seed`.

sobolevn: https://docs.pytest.org/en/latest/doctest.html#the-doctest-namespace-fixture

lk-geimfari: @sobolevn But what if we update data source? Will result change?

sobolevn: I am not sure. Possibly. But in this case it is wise to double check the failing docs. Since it might list the removed data.

lk-geimfari: Okay. So, I'll update all the doctets tonight. Thanks!

lk-geimfari: @sobolevn I have tried to write doctests with seed and as I can judge it's really unreliable solution. So, I'm not sure what else can I do. I'll remove all the doctets to work on them in a different branch. Right now I'll work only on refactoring and improving code, not doctests.

PR #547, opened 2018-10-22
'choice': tweak & make a provider

### LINKED ISSUE #484: Support `choice` provider
I want to be able to select any value from a given sequence. 

Like:

```python
mimesis('multiple_choice', items=['a', 'b', 'c'], number=2)
# => ['b', 'a']

mimesis('multiple_choice', items=['a', 'b', 'c'])
# => ['b']

mimesis('choice', items=['a', 'b', 'c'])
# => 'a'
```

It seems to be imposible right now.

sadielbartholomew:
**Close #484 by converting ``choice`` to a provider**. Additionally:

* **Change behaviour (& extend testing accordingly) for inconsistent behaviour** noticed but not picked up by tests when setting ``unique=True``. The number of elements (on master) provided was different to (less than) ``number`` if there were insufficient unique elements in ``items``:
    ```python
    >>> choice(['a', 'b', 'c'], number=4, unique=True)
    ['b', 'c']
    >>> choice(['a', 'b', 'c'], number=4, unique=True)
    ['c', 'a', 'b']
    ```
    Now raise an error if there are not enough unique elements to satisfy ``number`` (now ``length`` - see 
    below)

* **Change the interface slightly** to be more intuitive: in #484 there was the distinction between 'multiple choice', returning a list, & 'choice', returning an uncontained element. In this implementation, ``number`` is renamed ``length`` implying a **sequence length**. The user would like a sequence. So if specified (& valid i.e. a positive integer), a sequence is returned, but if unspecified, just return a *bare* element i.e. one not in a sequence. Prevent input of empty sequences, as making a random choice over these makes little sense.
* **Extend** the functionality so that the output for other sequences that make sense in context (i.e. ``tuple`` & ``str``) will be of the same type as the input ``items``, not always a list.
* **Document** ``choice``, which was previously missing from the docs. Interactive shell examples are included in the API reference (copied for illustrative convenience here):
    ```python
    >>> from mimesis import Choice
    >>> choice = Choice()
    
    >>> choice(items=['a', 'b', 'c'])
    'c'
    >>> choice(items=['a', 'b', 'c'], length=1)
    ['a']
    >>> choice(items='abc', length=2)
    'ba'
    >>> choice(items=('a', 'b', 'c'), length=5)
    ('c', 'a', 'a', 'b', 'c')
    >>> choice(items='aabbbccccddddd', length=4, unique=True)
   'cdba'
    ```


mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+
+from mimesis.providers.base import BaseDataProvider
+
+__all__ = ['Choice']
+
+
+class Choice(BaseDataProvider):
+    """Class for generating a random choice from items in a sequence."""
+
+    def __init__(self, *args, **kwargs):
```
lk-geimfari: It's a good idea to annotate it: `def __init__(self, *args, **kwargs) -> None:`

mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+
+        :param items: Non-empty sequence (list, tuple or string) of elements.
+        :param length: Length of sequence (number of elements) to provide.
+        :param unique: If True, ensures provided elements are unique.
+        :return: Sequence or uncontained element randomly chosen from items.
+        :raises TypeError: For non-sequence items or non-integer length.
+        :raises ValueError: If negative length or insufficient unique elements.
+
+        :Example:
+            ['b', 'c', 'a']
```
lk-geimfari: We don't need to keep both examples, please, remove it (['b', 'c', 'a']).

mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+        'c'
+        >>> choice(item=['a', 'b', 'c'], length=1)
+        ['a']
+        >>> choice(item='abc', length=2)
+        'ba'
+        >>> choice(item=('a', 'b', 'c'), length=5)
+        ('c', 'a', 'a', 'b', 'c')
+        >>> choice(item='aabbbccccddddd', length=4, unique=True)
+        'cdba'
+
```
lk-geimfari: Extra white space. Let's remove it!

mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+        elif len(items) == 0 or length < 0:
+            raise ValueError('**Items** must be a non-empty sequence and '
+                             '**number** should be a positive integer.')
+
+        if length == 0:  # else hereafter number >= 1 from constraints above
+            return self.random.choice(items)
+
+        # Process result as list then convert back to original type if non-list
+        data = []  # type: ignore
+        if unique and len(set(items)) < length:  # avoid an infinite while loop
```
lk-geimfari: Maybe just return as many unique as we have in sequence?

mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+            raise ValueError('There are not enough unique elements in '
+                             '**items** to provide the specified **number**.')
+        while len(data) < length:
```
lk-geimfari: We can make sequence unique just like that:

```python
data = []
while len(data) < length:
    item = self.random.choice(items)
    data.append(item)

if unique:
    data = list(set(data))

```

mimesis/providers/choice.py
```diff
@@ -0,0 +1,83 @@
...
+            item = self.random.choice(items)
+            if (unique and item not in data) or not unique:
+                data.append(item)
+        if isinstance(items, list):
+            return data
+        elif isinstance(items, tuple):
+            return tuple(data)
+        else:  # string is the only remaining possibility
```
lk-geimfari: Here no need in this comment. It's obvious.

lk-geimfari:
@sobolevn Do you have something to add to my comments?

sadielbartholomew:
Thank you for the quick & helpful feedback @lk-geimfari! I've addressed and squashed into the one commit the trivial parts.

I am not sure about the ``unique`` parameter feedback suggestion.

> Maybe just return as many unique as we have in sequence?

For cases where there are fewer unique elements in ``items`` than the ``length`` (or ``number`` before) this would seem not correct (inconsistent) behaviour to me as ``length`` would be different to the provided sequence length, as demonstrated in my initial comment examples. Though another option is to maybe change that variable name to ``max_length``? Else there is the way I implemented to raise an error for inconsistency.

Can we get a second opinion from reviewer 2? I will happily change the code to implement whatever you both think is best with respect to ``unique``, though (just maybe get another perspective first).

lk-geimfari:
@sadielbartholomew Sure! @sobolevn 🏓

lk-geimfari:
@sadielbartholomew So, let's keep your solution (I'm about `unique`)!

lk-geimfari:
@sadielbartholomew Do you checked other my comments, can we merge this PR?

sadielbartholomew:
@lk-geimfari sorry for the delay, I had to wait until my lunch break. Yes, this is ready for merging if you are both happy. Thanks!

lk-geimfari:
@sadielbartholomew Great! Thank you!

PR #563, opened 2018-11-05
Fix #562

### LINKED ISSUE #562: Locale overrider support
# Feature request

## Thesis

So, the idea is very simple - use context manager to temporarily override current locale for provider's instance, just like that:

```python
>>> from mimesis import Person
>>> from mimesis import locales

>>> person = Person(locales.EN)
>>> person.full_name()
'Ozie Melton'

>>> with locales.override(person, locales.RU):
...     person.full_name()

'Симона Богданова'

>>> person.full_name()
'Waldo Foster'
```

## Reasoning

Sometimes we need only some data from other provider and creating an instance each time is not really good, so it's better just override locale.

lk-geimfari:
# I have made things!

## Checklist

## Related issues

- Closes #562

### DIFF
--- a/mimesis/locales.py
+++ b/mimesis/locales.py
@@ -1,3 +1,5 @@
+from mimesis.utils import pull
+
 CS = 'cs'
 DA = 'da'
 DE = 'de'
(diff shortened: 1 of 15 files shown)


mimesis/locales.py
```diff
@@ -31,3 +33,35 @@
 TR = 'tr'
 UK = 'uk'
 ZH = 'zh'
+
+
+class override(object):  # noqa
```
sobolevn: Why not a generator?

mimesis/locales.py
```diff
@@ -31,3 +33,35 @@
...
+    """Context manager which allows overriding current locale.
+
+     .. note:: This feature does not works with :class:`~mimesis.Generic()`.
+        Usually, you don't need data from other locales,
+        when you use :class:`~mimesis.Generic()`.
+    """
+
+    def __init__(self, provider, locale: str = EN) -> None:
```
sobolevn: `provider` can be also typed.

mimesis/locales.py
sobolevn: Yeap

mimesis/locales.py
```diff
@@ -31,3 +37,40 @@
 TR = 'tr'
 UK = 'uk'
 ZH = 'zh'
+
+_LocaleDependProvider = typing.Union[
```
lk-geimfari: Not sure about this, but this is a temporary solution until we close #565.

mimesis/locales.py
```diff
@@ -35,33 +37,28 @@
 ZH = 'zh'
 
 
-class override(object):  # noqa
+@contextlib.contextmanager
+def override(provider, locale: str = EN):
```
sobolevn: `-> Generator[Provider, None, None]`

PR #566, opened 2018-11-06
Fix #565

### LINKED ISSUE #565: Different superclasses for different types of providers
## Thesis

The providers which have locale dependent must be inherited from `BaseDataProvider`, but providers which just works with standard data structures, without storing data in JSON from `BaseProvider`.

## Reasoning
Locale independent providers must not contain any attributes of providers which has locale dependent. It can help make API much clear without the chaos of inheritance.

lk-geimfari:
# I have made things!

## Checklist

## Related issues

Format is:
- Closes #565

## Other significant changes.

- Now you can add to Generic only providers which are subclasses of ``BaseProvider`` to ensure a single instance of ``random.Random()`` for all providers.
- Updated documentation.
- Annotated `.add_providers()` method of `Generic()`

### DIFF
--- a/mimesis/providers/generic.py
+++ b/mimesis/providers/generic.py
@@ -48,8 +48,8 @@ class Generic(BaseDataProvider):
         self._text = Text
         self._food = Food
         self._science = Science
-        self._code = Code
-        self._transport = Transport
+        self.transport = Transport(seed=self.seed)
+        self.code = Code(seed=self.seed)
         self.unit_system = UnitSystem(seed=self.seed)
         self.file = File(seed=self.seed)
         self.numbers = Numbers(seed=self.seed)
--- a/mimesis/providers/generic.py
+++ b/mimesis/providers/generic.py
@@ -1,10 +1,10 @@
 """Provides all at one."""
 
 import inspect
-from typing import Any, Callable, List
+from typing import Any, List, Sequence
 
 from mimesis.providers.address import Address
-from mimesis.providers.base import BaseDataProvider
+from mimesis.providers.base import BaseDataProvider, BaseProvider
 from mimesis.providers.business import Business
 from mimesis.providers.choice import Choice
 from mimesis.providers.clothing import ClothingSize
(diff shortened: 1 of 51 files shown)


mimesis/providers/generic.py
```diff
@@ -104,17 +103,21 @@ def add_provider(self, cls) -> None:
         :raises TypeError: if cls is not class.
         """
         if inspect.isclass(cls):
```
sobolevn: We need to annotate this method, it should be `cls: Type[BaseProvider]`.

mimesis/providers/generic.py
```diff
@@ -104,17 +103,21 @@ def add_provider(self, cls) -> None:
...
                     name = cls.Meta.name
             else:
                 name = cls.__name__.lower()
+
             setattr(self, name, cls())
         else:
             raise TypeError('Provider must be a class')
 
-    def add_providers(self, *providers) -> None:
+    def add_providers(self, *providers: Sequence[BaseProvider]) -> None:
```
sobolevn: Are you sure that this is not `Sequence[Type[BaseProvider]]`?

mimesis/providers/transport.py
```diff
@@ -56,14 +58,13 @@ def airplane(self, model_mask: str = '###') -> str:
...
         """Get vehicle registration code of country.
 
-        :param allow_random: If False return only code
-            for current locale (country).
+        :param locale: Registration code for locale (country).
         :return: Vehicle registration code.
         """
-        if not allow_random:
-            return VRC_BY_LOCALES[self.locale]
+        if locale:
```
sobolevn: Why do we need this change? What's your motivation?

lk-geimfari: To make `Code` locale-independent, because only this method needs locale code in this provider. So, it better just pass locale code to this method, instead of making this provider locale-dependent.

lk-geimfari:
That's done!

Issue #570, opened 2018-11-10
Rename modules with bad names.


lk-geimfari:
Read more here #568 

Modules:

- `helpers.py`
- `__version__.py`
- `utils.py`

lk-geimfari:
It's done!

Issue #579, opened 2018-11-18
Use crypto-secure random module


sobolevn:
# Feature request

## Thesis

I would like to switch from `random` module to a secure version of random module: `secrets`. 
It is available since `python3.6` and is cryptographically secure.

It has the same API as regular `random`, so the switch should be painless.

## Reasoning

Having this feature is a good thing. Some users might generate data that is used in real world apps, so we need to take care of them for this usecase.

lk-geimfari:
@sobolevn Do you mean this?

```python
from secrets import SystemRandom
```

sobolevn:
Yes, like this.

But, I need to mention that we need to fallback to `random.Random`.
Because 

```python
class SystemRandom(Random):
    """Alternate random number generator using sources provided
    by the operating system (such as /dev/urandom on Unix or
    CryptGenRandom on Windows).

     Not available on all systems (see os.urandom() for details).
    """
```

lk-geimfari:
@sobolevn So, should I close this issue? Is there anything else we can do?

sobolevn:
I am a little bit lost. I guess that it was never implemented. Wasn't it?

lk-geimfari:
Unfortunately wasn't. I can work on it, but I'm not sure what should I do.

sobolevn:
Ok, let's close it. I don't remember what was my original intention.

sobolevn:
P.S. If you want to work on some open-source - we have a lot of opened tasks in `wemake-python-styleguide`: https://github.com/wemake-services/wemake-python-styleguide/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22 I would appreciate your help.

lk-geimfari:
I'll try to work on some issue. Thanks!

PR #583, opened 2018-11-20
Refactoring

### LINKED ISSUE #568: Moving utils.pull to BaseDataProvider
# Feature request

## Thesis
I suggest moving `utils.pull` to `BaseDataProvider` because it's only needed for locale-dependent providers, which are all subclasses of `BaseDataProvider`.

## Reasoning
We are forced to always manually import `pull` to the modules where are we create new providers and repeat the same code, although the pull is needed only for locale-dependent providers which are all inherited from `BaseDataProvider`. It will be much logically correct to keep pull in provider `BaseDataProvider` I think.

The idea is simple, instead of this:
```python
from mimesis.utils import pull

class OldProvider(BaseDataProvider):

    def __init__(self, *args, **kwargs):
        """Initialize attributes.
        """
        super().__init__(*args, **kwargs)
        self._datafile = 'datetime.json'
        self._data = pull(self._datafile, self.locale)
```

Use this:

```python
class NewProvider(BaseDataProvider):

    def __init__(self, *args, **kwargs):
        """Initialize attributes.
        """
        super().__init__(*args, **kwargs)
        self._datafile = 'datetime.json'
        self.pull(self._datafile)
```
After this we'll able to use pull withot any extra importing.

That's just discussion, I have not started working on this yet.

### LINKED ISSUE #569: Merge modules config and locales
# Feature request

## Thesis
I suggest merging this two modules (`config.py` and `locales.py`). The module `confing.py` contains only things related to locale, so it's better to keep only module `locales.py` and move all the content of `config.py` to `locales.py`.

## Reasoning

Here is no need in two different modules for the same things.

lk-geimfari:
# I have made things!

## Checklist

- [x] Merge `config.py` and` locales.py`
- [x] Move pull to `BaseDataProvider`
- [x] Update locales.override to support a new way of data pulling from files.
- [x] Delete module `utils.py`
- [x] Rewrite tests for moved functions

## Related issues

- Closes #568, #569, #570


lk-geimfari:
I know that' really hard to review such PRs, but there is no another way.

lk-geimfari:
@sobolevn I need your review again and after this time I'll merge this PR if you have not any pieces of advice/notices.

lk-geimfari:
I did not fix this https://github.com/lk-geimfari/mimesis/issues/579, because I'm not sure how to do it.

mimesis/providers/base.py
```diff
@@ -71,16 +71,16 @@ def __str__(self) -> str:
...
+                 seed: Seed = None) -> None:
         """Initialize attributes for data providers.
 
         :param locale: Current locale.
         :param seed: Seed to all the random functions.
         """
         super().__init__(seed=seed)
         self._data: dict = {}
-        self._datafile: Optional[str] = None
+        self._datafile: str = ''
```
sobolevn: Useless annotation

mimesis/providers/base.py
```diff
@@ -71,16 +71,16 @@ def __str__(self) -> str:
...
-                 seed: Optional[Seed] = None) -> None:
+    def __init__(self, locale: str = locales.DEFAULT_LOCALE,
+                 seed: Seed = None) -> None:
         """Initialize attributes for data providers.
 
         :param locale: Current locale.
         :param seed: Seed to all the random functions.
         """
         super().__init__(seed=seed)
         self._data: dict = {}
```
sobolevn: Should be annotated as: `Dict[KeyType, ValueType`. `dict` means `Dict[Any, Any]`. And it does not help much.

lk-geimfari: @sobolevn The problem is that we don't know `ValueType`.

sobolevn: It may be `Any` in this case.

sobolevn: It might be useful to enable `disallow_any_generics = True`

Issue #561, opened 2018-11-05
Add support of parameter unique for email, username and telephone.


lk-geimfari:
# Feature request

## Thesis
This feature requested by @neumond

He propose to add parameter `unique` for `.email()`, `.username()` and `.telephone()` to make sure that they return only unique data while the provider's instance exists.

## Reasoning
Some fields must be unique and it would be a good idea to add support of parameter `unique` for such methods (`.email()`, `.username()` and `.telephone()`).

jackhenry:
Hi @lk-geimfari 

I think this is a good idea and would be interested in implementing this.

One thing I want to mention is that during the life of the provider's instance there are a finite amount of unique values it could return. For example, when generating a ten digit phone number there are 10<sup>10</sup> unique possible phone numbers.

For ```.email()``` and ```.email()```, the amount of unique values is directly related to the amount of values in ```person.json```

Even though it's unlikely a single provider would be used to try to generate all possible unique values, should the implementation take into account that there are a finite amount of unique values possible for  ```.email()``` ```.username()``` and ```.telephone()```?

PR #608, opened 2019-01-01
[WIP] Unique parameter for Person.email()

jackhenry:
# I have made things!

## Checklist

- [ ] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

## Related issues

- Refs #561 

This my first pull request so I apologize for any mistakes.

So far, I have only implemented the unique parameter for ```Person.email()```. I wanted my current progress to be reviewed before I continue any further. Thanks.

### DIFF
--- a/mimesis/providers/person.py
+++ b/mimesis/providers/person.py
@@ -180,7 +180,9 @@ class Person(BaseDataProvider):
         _name = self.random.choice(USERNAMES).capitalize()
 
         name = self.random.choice(USERNAMES)
-        date = self.random.randint(1800, 2070)
+        MIN_DATE = 1800
+        MAX_DATE = 2070
+        date = self.random.randint(MIN_DATE, MAX_DATE)
 
         templates = ('U-d', 'U.d', 'UU-d', 'UU.d', 'UU_d', 'U_d',
                      'Ud', 'l-d', 'l.d', 'l_d', 'ld', 'default')
--- a/mimesis/providers/person.py
+++ b/mimesis/providers/person.py
@@ -249,7 +249,7 @@ class Person(BaseDataProvider):
 
         :param domains: List of custom domains for emails.
         :type domains: list or tuple
-        :param unique: Flag to generate a unique email
+        :param unique: Flag to generate a unique email.
         :type unique: bool
         :return: Email address.
 
--- a/mimesis/providers/person.py
+++ b/mimesis/providers/person.py
@@ -2,7 +2,7 @@
 
 import hashlib
 from string import ascii_letters, digits, punctuation
-from typing import Optional, Union
+from typing import Dict, Optional, Set, Union
 
 from mimesis.data import (
     BLOOD_GROUPS,
@@ -35,6 +35,9 @@ class Person(BaseDataProvider):
         self.pull(self._datafile)
         self._store = {
             'age': 0,
+        }
+        self._previously_generated: Dict[str, Set] = {
+            'emails': set(),
         }
 
     def age(self, minimum: int = 16, maximum: int = 66) -> int:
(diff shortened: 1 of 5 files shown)


mimesis/providers/person.py
```diff
@@ -36,6 +36,9 @@ def __init__(self, *args, **kwargs) -> None:
         self._store = {
             'age': 0,
         }
+        self._previously_generated: Dict[str, Set] = {
```
lk-geimfari: Why do not use `self._store`? I guess `self._store` has been created for such things when you need to know previously generated values. What do you think? It's important for us to know your point, so feel free to oppose.

mimesis/providers/person.py
```diff
@@ -238,11 +241,14 @@ def password(self, length: int = 8, hashed: bool = False) -> str:
...
 
-    def email(self, domains: Union[tuple, list] = None) -> str:
+    def email(self, domains: Union[tuple, list] = None,
+              unique: bool = False) -> str:
         """Generate a random email.
 
         :param domains: List of custom domains for emails.
         :type domains: list or tuple
+        :param unique: Flag to generate a unique email.
+        :type unique: bool
```
lk-geimfari: Actually, we do not need to add type to the docstrings, because we use sphinx-type-hints, so you can remove lines starts with `type:`.

mimesis/providers/person.py
```diff
@@ -253,11 +259,24 @@ def email(self, domains: Union[tuple, list] = None) -> str:
...
+        if unique:
+            while email in self._previously_generated['emails']:
+                domain = self.random.choice(domains)
+                name = self.username(template='ld')
+                email = '{name}{domain}'.format(
+                    name=name,
+                    domain=domain,
+                )
+
+        self._previously_generated['emails'].add(email)
```
lk-geimfari: As I say above, I think it will better to use `self._store` here, like this: `self._stor['emails'].add(email)`, but if you do not agree, we can discuss it.

lk-geimfari:
@jackhenry Hey! Firstly, happy new year. Secondly, I have added a little review.

I have a very important question:  what will we do when we exhaust unique values?

jackhenry:
@lk-geimfari Happy New Year to you as well! I responded to your comments in the review.

Good question. I have been thinking of ways to deal with exhausted unique values and would love your input. 

For ```.email(unique=True)```, the amount of unique values possible is a combination of all valid usernames and all valid domains. I was thinking that when a ```Person()``` instance is created, the total possible unique values is computed by using the formula for computing all possible combinations. (My stats is rusty so I still need to research what formula to use)

Once the number of all possible unique values is generated, ```.email()``` will increment a counter every time a new unique value is returned. If that counter reaches the amount of all possible values, a couple things can happen.

**1**. Some exception will be thrown.
or
**2**. The counter will reset and the set of all previously generated values will be cleared.

I personally like option 1 more.

Let me know what you think.

lk-geimfari:
@jackhenry I also like the first option. Throwing exception is much better.

mimesis/providers/person.py
jackhenry: Noted. I will remove type from the docstrings.

mimesis/providers/person.py
jackhenry: I originally used ```self._store```. However, adding a new key value pair with type ```[str, Set]``` would result in a dictionary with multiple types of key value pairs. I wasn't sure if this was the cleanest way to do this.

Also note, when I implement ```Person.username(unique=True)``` and ```Person.telephone(unique=True)``` I will also be adding more ```[str, Set]``` key value pairs. Overall, I thought it would be cleaner to have a dedicated single type dictionary. It would only be used for keeping track of previously generated values for all the methods that implement ```unique=True```.

However, if you think I should just use ```self._store``` I will happily revise the code.

lk-geimfari: Great!

lk-geimfari:
@jackhenry 🏓

ceccoemi:
Hi, it seems that this pull request is stuck.

@lk-geimfari do you mind if I try to write from scratch this feature in a new branch?

lk-geimfari:
@ceccoemi Of course not. You can try to do it! 👍

PR #611, opened 2019-01-04
Refine Person.username()

jackhenry:
# I have made things!

## Checklist

- [ ] I have created at least one test case for the changes I have made

## Related issues

Hi @lk-geimfari,

I'm still working hard on #561 and #608. While implementing the unique parameter for ```Person.username()```, I noticed this function could be refined. 

Let me know what you think.

### DIFF
--- a/mimesis/providers/person.py
+++ b/mimesis/providers/person.py
@@ -1,6 +1,7 @@
 """Provides personal data."""
 
 import hashlib
+import re
 from string import ascii_letters, digits, punctuation
 from typing import Optional, Union
 
@@ -176,19 +177,18 @@ class Person(BaseDataProvider):
         :Example:
             Celloid1873
         """
-        # TODO: Optimize
-        _name = self.random.choice(USERNAMES).capitalize()
-
-        name = self.random.choice(USERNAMES)
         MIN_DATE = 1800
         MAX_DATE = 2070
-        date = self.random.randint(MIN_DATE, MAX_DATE)
 
-        templates = ('U-d', 'U.d', 'UU-d', 'UU.d', 'UU_d', 'U_d',
-                     'Ud', 'l-d', 'l.d', 'l_d', 'ld', 'default')
+        templates = ('U_d', 'U.d', 'U-d', 'UU-d', 'UU.d', 'UU_d',
+                     'ld', 'l-d', 'Ud', 'l.d', 'l_d', 'default')
+        DEFAULT_TEMPLATE = 'l.d'
 
         if template is None:
             template = self.random.choice(templates)
+
+        if template is 'default':
+            template = DEFAULT_TEMPLATE
 
         if template not in templates:
             raise KeyError(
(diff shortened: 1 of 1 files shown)


mimesis/providers/person.py
```diff
@@ -176,49 +177,43 @@ def username(self, template: Optional[str] = None) -> str:
...
-        name = self.random.choice(USERNAMES)
         MIN_DATE = 1800
         MAX_DATE = 2070
-        date = self.random.randint(MIN_DATE, MAX_DATE)
 
-        templates = ('U-d', 'U.d', 'UU-d', 'UU.d', 'UU_d', 'U_d',
-                     'Ud', 'l-d', 'l.d', 'l_d', 'ld', 'default')
+        templates = ('U_d', 'U.d', 'U-d', 'UU-d', 'UU.d', 'UU_d',
+                     'ld', 'l-d', 'Ud', 'l.d', 'l_d', 'default')
+        DEFAULT_TEMPLATE = 'l.d'
```
lk-geimfari: Move this above `templates`, please. Like this:

```
        MIN_DATE = 1800
        MAX_DATE = 2070
        DEFAULT_TEMPLATE = 'l.d'

        templates = ('U_d', 'U.d', 'U-d', 'UU-d', 'UU.d', 'UU_d',
                     'ld', 'l-d', 'Ud', 'l.d', 'l_d', 'default')
```

mimesis/providers/person.py
```diff
@@ -176,49 +177,43 @@ def username(self, template: Optional[str] = None) -> str:
...
 
         if template is None:
             template = self.random.choice(templates)
 
+        if template is 'default':
```
lk-geimfari: if template == 'default', not `is`.

lk-geimfari:
@jackhenry Did you update the tests? Are they pass?

jackhenry:
@lk-geimfari All tests pass. I don't think the test needs to change because the test still covers all possible patterns ```username()``` can return. 

However, ```username()``` could support any custom template if the following code was removed.
```
if template not in templates:
    raise KeyError
```
Then I would have to rewrite the test and add a check to make sure the passed template only contained valid characters. Although, I'm not really sure if that's a feature you are interested in.

mimesis/providers/person.py
```diff
@@ -176,49 +177,43 @@ def username(self, template: Optional[str] = None) -> str:
...
+            template = DEFAULT_TEMPLATE
+
         if template not in templates:
             raise KeyError(
                 'Template «{template}» is not in {templates}'.format(
```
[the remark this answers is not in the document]
jackhenry: I agree.

mimesis/providers/person.py
jackhenry: Oops.

mimesis/providers/person.py
jackhenry: Will do.

lk-geimfari:
@jackhenry Don't you think that this may complicate your work? If not, you can rewrite without `raise`.

lk-geimfari:
@jackhenry If you will rewrite it or extend possible templates don't forget to explain how it will work in docstrings.

jackhenry:
@lk-geimfari It would definitely complicate the work. Instead of computing the possible combinations for a finite amount of username templates, it would have to be computed on the fly depending on what template is passed to the function.

Supporting custom username templates probably isn't a feature that is super important to the users of this library. I think the current supported templates are adequate. However, I'll let you have the final say.

lk-geimfari:
@jackhenry I agree. We do not need such complexity.

lk-geimfari:
@jackhenry Can I merge this PR?

PR #640, opened 2019-02-28
Refactor Path data provider with pathlib

### LINKED ISSUE #607: Use pathlib for Path data provider
It would be a good idea to use [pathlib](https://docs.python.org/3/library/pathlib.html) for [Path](https://mimesis.name/api.html#path) because current implementation is too inflexible (we're manually checking operation system on generating paths).

ceccoemi:
# I have made things!

## Checklist

## Related issues

- Closes #607
- Refs #607 

I've refactored the Path data provider using the [pathlib](https://docs.python.org/3/library/pathlib.html) library as suggested in #607.

In order to remove that strange loops over the `PLATFORMS` dictionary I have changed the internals of the `path.py` module: now the main keys are the platform strings (*linux*, *darwin*, *win32* and *win64*) and the only thing specified in the dictionary is the *home* path:

```
PLATFORMS = {
    'linux': {
        'home': '/home/',
    },
    'darwin': {
        'home': '/home/',
    },
    'win32': {
        'home': 'C:\\Users\\',
    },
    'win64': {
        'home': 'C:\\Users\\',
    },
}
```

I have introduced a private variable `_pathlib_home` that is a [PurePath](https://docs.python.org/3/library/pathlib.html?highlight=purepath#pathlib.PurePath) object ([PureWindowsPath](https://docs.python.org/3/library/pathlib.html?highlight=purepath#pathlib.PureWindowsPath) or [PurePosixPath](https://docs.python.org/3/library/pathlib.html?highlight=purepath#pathlib.PurePosixPath) depending on the specified platform). That object contains the path to the home directory, all the other paths are built from the *home* path, for instance:

```
def root(self) -> str:
    """Generate a root dir path.
    :return: Root dir.
    :Example:
        /
        """
        return str(self._pathlib_home.parent)

...

def user(self) -> str:
    """Generate a random user.
    :return: Path to user.
    :Example:
        /home/oretha
    """
    user = self.random.choice(USERNAMES)
    user = user.capitalize() if 'win' in self.platform else user.lower()
    return str(self._pathlib_home / user)
```

In the refactoring process I've noticed that the unit tests for the `path` module were wrong: I fixed the assert statements to have actual testing (see the second commit of this pull request).

### DIFF
--- a/mimesis/data/int/path.py
+++ b/mimesis/data/int/path.py
@@ -4,28 +4,16 @@
 
 # https://docs.python.org/3.5/library/sys.html?highlight=sys#sys.platform
 PLATFORMS = {
-    'LINUX': {
-        'name': 'linux',
-        'path_separator': '/',
-        'root': '/',
+    'linux': {
         'home': '/home/',
     },
-    'MACOS': {
-        'name': 'darwin',
-        'path_separator': '/',
-        'root': '/',
+    'darwin': {
         'home': '/home/',
     },
-    'WINDOWS': {
-        'name': 'win32',
-        'path_separator': '\\',
-        'root': 'C:\\',
+    'win32': {
         'home': 'C:\\Users\\',
     },
-    'WINDOWS64': {
-        'name': 'win64',
-        'path_separator': '\\',
-        'root': 'C:\\',
+    'win64': {
         'home': 'C:\\Users\\',
     },
 }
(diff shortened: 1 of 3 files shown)


lk-geimfari:
@ceccoemi One more thing. How about adding yourself to `CONTRIBUTORS.rst`?

lk-geimfari:
Oh, I see that you have done it in the previous PR.

PR #655, opened 2019-03-21
Unique parameter

### LINKED ISSUE #561: Add support of parameter unique for email, username and telephone.

ceccoemi:
# I have made things!

## Related issues

Format is:
- Closes #561 
- Refs #561

I have added the `unique` parameter to the functions `username()`, `email()` and `telephone()`.

Before merge this pull request ... @lk-geimfari do you think that there are other functions (also from other providers) that need the unique parameter?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ Version 3.2.0
 
 - Added built-in provider DenmarkSpecProvider
 - Added meta classes for providers for internal usage (see `#621 <https://github.com/lk-geimfari/mimesis/issues/621>`_.)
+- Added unique parameter for some functions of the Person provider
 
 **Fixed**:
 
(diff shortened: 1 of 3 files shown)


lk-geimfari:
@ceccoemi Current implementation does not cover cases when unique values exhausted. This problem is more complicated than you think, please take a look at comments in #608.

It must return only unique values and raise an exception when such values exhausted.

ceccoemi:
@lk-geimfari I misunderstood the specifications. I'm sorry, I wil continue to work on it.

PR #658, opened 2019-03-27
Username dev

ceccoemi:
# I have made things!

## Checklist

## Related issues

The function `username()` in the `Person` provider now support custom templates. The specifications of the template format are written in the doctstring of the function.

I tried not to break the compatibility with the old function, so that if `template=None` or `template='default'` then the returned username is the same as before.

@lk-geimfari what do you think about replacing the type of the exception? I think that an `IndexError` exception is not suitable for a wrong template format, what about throwing a `ValueError` instead?


tests/test_providers/test_person.py
```diff
@@ -88,17 +85,22 @@ def test_username(self, _person, template):
             'l_d': r'^[a-z]+_[0-9]+$',
             'ld': r'^[a-z]+[0-9]+$',
             'default': r'^[a-z]+\.[0-9]+',
+            None: r'^[A-Za-z]{2,}[\.\-\_]?[0-9]+$',
```
lk-geimfari: @ceccoemi Are you sure that tests cover all the templates?

mimesis/providers/person.py
```diff
@@ -184,28 +191,21 @@ def username(self, template: Optional[str] = None) -> str:
         if template == 'default':
             template = DEFAULT_TEMPLATE
 
-        if template not in templates:
+        if re.fullmatch(r'[Ul\.\-\_d]*[Ul]+[Ul\.\-\_d]*', template) is None:
```
lk-geimfari: @ceccoemi Can we write it like this?

```python
if not re.fullmatch(r'[Ul\.\-\_d]*[Ul]+[Ul\.\-\_d]*', template):
    ... 

```

tests/test_providers/test_person.py
ceccoemi: Yes I have spent a lot of time in building that regex and I am pretty sure that it covers all the templates that are used when `template=None`.

I want to clarify that the regex do not cover all the possible usernames with all the possible templates, but it refers to `(‘U_d’, ‘U.d’, ‘U-d’, ‘UU-d’, ‘UU.d’, ‘UU_d’, ‘ld’, ‘l-d’, ‘Ud’, ‘l.d’, ‘l_d’)`, that are the templates used when `template=None`.

mimesis/providers/person.py
ceccoemi: Sure! I will fix this

lk-geimfari:
@ceccoemi You can merge it.

lk-geimfari:
@ceccoemi About exception: Sure, you can replace it with much suitable. Don't forget to update docstrings and tests.

ceccoemi:
@lk-geimfari the pull request is ready. Due to my last commit, you have to approve it again.

lk-geimfari:
@ceccoemi You can merge it! 👍

PR #697, opened 2019-06-26
Country locale

### LINKED ISSUE #696: Country generation does not respect chosen locale
It appears that `Address('en').country()` does not respect the locale. Countries outside locale are still generated. 

Looking at the code it appears there is no implementation for locale specific country, if so, please change issue to a feature request.

ceccoemi:
# I have made things!

## Checklist

## Related issues

Format is:
- Closes #696 
- Refs #696

Now the `country()` function in the `Address` provider by default return the name of the country of the current locale:

    >>> a = Address(locale='en')
    >>> a.country()
    'United States'
    >>> a.country()
    'United States'

To get a random country you must set the `allow_random` parameter to `True`:

    >>> a = Address(locale='en')
    >>> a.country(allow_random=True)
    'Japan'
    >>> a.country(allow_random=True)
    'France'


lk-geimfari:
@ceccoemi LGTM. You can merge it when all tests will pass.

ZeroCool940711:
I know this is a few months old but would be nice if `address.state()` and `address.city()` worked like `address.country()`, I mean, if your locale is EN you should be getting addresses for the United States only but also if the state is Arizona you should get only cities from Arizona and not from other states, also the change to Address that allow you to use a locale only works when you use directly `mimesis.Address` but doesnt work if you use it from `mimesis.Generic.address`, another thing related to this is phone numbers, we should be able to get phone numbers for a specific State or City, all those things should work together to create better results.

lk-geimfari:
@ZeroCool940711 Such features must be implemented in `USASpecProvider`. You can create PR with your changes.

ZeroCool940711:
@lk-geimfari I don't know enough python or how mimesis works under the hood and don't have enough free time to implement that myself, would appreciate if someone else could do that.

PR #714, opened 2019-08-15
Add Windows support to Travis CI

lk-geimfari:
Now [Windows is available on Travis CI](https://blog.travis-ci.com/2018-10-11-windows-early-release), so we can reject using AppVeyor.


sobolevn:
WOW 🛒

PR #767, opened 2019-11-29
Matrix

### LINKED ISSUE #727: The matrix generators for provider Numbers()
# Feature request

## Thesis
The matrix can be useful in working in Jupyter Notebook. So, it would be awesome to support generators of matrix in provider `Numbers()`.

I sketched a rough implementation example:

```python
import random

from typing import Tuple, Optional, Union, List

ValuesRange = Union[
    List[int],
    Tuple[int, int],
]


def matrix(m: int = 3, n: int = 3, values_range: Optional[ValuesRange] = None) -> list:
    """Creates an 𝑚×𝑛 matrix which has 𝑚 rows and 𝑛 columns.

    :param m: Rows count.
    :param n: Columns count.
    :param values_range: Range of values of matrix items.
    :return: Matrix 𝑚×𝑛.
    """
    if values_range is None:
        values_range = (-100, 100)

    if len(values_range) > 2:
        raise ValueError('Invalid items range! Usage example: items_range=(-10, 10)')

    result = [
        [random.randint(*values_range) for _ in range(m)]
        for _ in range(n)
    ]

    return result

matrix(m=5, n=4, values_range=[-50, 50])
```

Output:
```python
[[29, 7, 17, -24, -29],
 [-26, 13, 29, -25, 6],
 [-5, 16, 13, -45, 49],
 [23, 20, -25, 14, -25]]
```

ceccoemi:
# I have made things!

## Checklist

## Related issues

Format is:
- Closes #727 
- Refs #727

I tried to follow the symmetry with the ``integers()`` function of the same provider, that's why I use the two parameters ``start`` and ``end``. The implementation is quite straightforward.

### DIFF
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -32,7 +32,8 @@ class Numbers(BaseProvider):
         """Generate a list of random integers.
 
         Integers can be negative or positive numbers.
-        .. note: You can use both positive and negative numbers.
+
+        note: You can use both positive and negative numbers.
 
         :param start: Start.
         :param end: End.
@@ -42,8 +43,7 @@ class Numbers(BaseProvider):
         :Example:
             [-20, -19, -18, -17]
         """
-        return self.random.randints(
-            length, start, end)
+        return self.random.randints(length, start, end)
 
     @staticmethod
     def primes(start: int = 1, end: int = 999) -> List[int]:
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -45,6 +45,21 @@ class Numbers(BaseProvider):
         """
         return self.random.randints(length, start, end)
 
+    def matrix(self, m: int = 3, n: int = 3,
+               start: int = 0, end: int = 10) -> list:
+        """Creates a m x n matrix of random integers.
+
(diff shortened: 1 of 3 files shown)


mimesis/providers/numbers.py
```diff
@@ -42,8 +43,22 @@ def integers(self, start: int = 0, end: int = 10,
...
+               start: int = 0, end: int = 10) -> list:
+        """Creates a m x n matrix of random integers.
+
+        :param m: Number of rows.
+        :param n: Number of columns.
+        :param start: Start range.
+        :param end: End range.
+        :return: Matrix with m rows and n columns.
+
+        :Example:
```
lk-geimfari: Don't you think that it will be much useful if we add support for the variety of types ( float, integer, complex etc.)?

ceccoemi: Yes, I was thinking the same thing. We should add an argument where we can specify the returned type.

Before implementing this feature, I would like to make some changes to this provider, please see #769

lk-geimfari:
@ceccoemi I'm closing this because of conflicts. Can you please recreate it?

ceccoemi:
@lk-geimfari I will recreate it as soon as I can.

PR #772, opened 2019-12-01
Numbers

### LINKED ISSUE #769: Restructure Numbers provider
# Feature request

## Thesis

While I was implementing the ``matrix()`` function in the ``Numbers`` provider and I was thinking about some changes that we could make in this provider:

- Add a function ``complex(start, end, length)`` that return a random array of complex numbers
- Make the API uniform, so that every function in the ``Numbers`` provider has the arguments ``start, end, length`` (where possible). Maybe in the ``complex()`` function we can add ``start_real, end_real, start_imaginary, end_imaginary`` ?
- Remove the function ``ranting()`` and add an argument ``decimal_digits`` in the function ``floats()`` to specify the number of decimal digits to keep.

## Reasoning

I think these changes would make the provider more uniform and easy to use.

ceccoemi:
# I have made things!

## Checklist

## Related issues

Format is:
- Closes #769 
- Refs #769 

Now you can specify the range of the float numbers generated and the maximum number of decimal digits.

I added the function `complexnums` (I didn't want to override the builtin `complex`), where you can specify the range for both real and imaginary part.

I removed the function `ranting`. It can be replaced with `floats(end=5, n=0, rounding=1)[0]`

### DIFF
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -122,14 +122,3 @@ class Numbers(BaseProvider):
         :return: Number.
         """
         return self.random.randint(minimum, maximum)
-
-    def rating(self, maximum: float = 5.0) -> float:
-        """Generate a random rating for something.
-
-        :param maximum: Maximum value (default is 5.0).
-        :return: Rating.
-
-        :Example:
-            4.7
-        """
-        return self.random.uniform(0, maximum, 1)
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -48,10 +48,10 @@ class Numbers(BaseProvider):
         return self.random.randints(
             length, start, end)
 
-    def complexnums(self, start_real: float = 0, end_real: float = 1,
-                    start_imag: float = 0, end_imag: float = 1,
-                    rounding_real: int = 15, rounding_imag: int = 15,
-                    length: int = 10) -> List[complex]:
+    def complexes(self, start_real: float = 0, end_real: float = 1,
+                  start_imag: float = 0, end_imag: float = 1,
+                  rounding_real: int = 15, rounding_imag: int = 15,
+                  length: int = 10) -> List[complex]:
         """Generate a list of random complex numbers.
 
         :param start_real: Start real range.
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -17,7 +17,7 @@ class Numbers(BaseProvider):
 
         name = 'numbers'
 
-    def floats(self, start: int = 0, end: int = 1, n: int = 2,
+    def floats(self, start: float = 0, end: float = 1, n: int = 2,
                rounding: int = 15) -> List[float]:
         """Generate a list of 10^n random float numbers.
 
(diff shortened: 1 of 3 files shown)


ceccoemi:
Windows build continues to fail :(

mimesis/providers/numbers.py
```diff
@@ -45,6 +48,26 @@ def integers(self, start: int = 0, end: int = 10,
         return self.random.randints(
             length, start, end)
 
+    def complexnums(self, start_real: float = 0, end_real: float = 1,
```
lk-geimfari: Don't you think that it's better to use similar names for methods? You know like floats, integers, complexes.

ceccoemi: Yes, I was thinking to `complexes()` too, but I wasn't sure it's a correct English word. I will change it.

lk-geimfari: Yeah, me too. I googled it: https://en.m.wiktionary.org/wiki/complex

lk-geimfari:
I think Windows fails because of `choco`.

mimesis/providers/numbers.py
```diff
@@ -17,15 +17,18 @@ class Meta:
...
+        """Generate a list of 10^n random float numbers.
 
+        :param start: Start range.
+        :param end: End range.
         :param n: Raise 10 to the 'n' power.
+        :param rounding: Max number of decimal digits.
         :return: The list of floating-point numbers.
         """
-        nums = [self.random.random()
+        return [self.random.uniform(start, end, rounding)
```
lk-geimfari: Why not just `n` random floats instead of 10^n?

ceccoemi: Agree, but I was trying to not break the API. Can I change the argument name to `length` or do you think it will break something?

lk-geimfari: Let it be n, but not 10^n. Do not forget about the changelog.

mimesis/providers/numbers.py
```diff
@@ -45,6 +48,26 @@ def integers(self, start: int = 0, end: int = 10,
         return self.random.randints(
             length, start, end)
 
+    def complexes(self, start_real: float = 0, end_real: float = 1,
+                  start_imag: float = 0, end_imag: float = 1,
+                  rounding_real: int = 15, rounding_imag: int = 15,
+                  length: int = 10) -> List[complex]:
```
lk-geimfari: We can use `n` there instead of `length` to make API similar to `floats`. What do you think? Should we?

ceccoemi: I like uniformity in the API, so I agree with changing `length`with `n` in every function.

lk-geimfari: Great! Thank you!

lk-geimfari:
@ceccoemi LGTM. You can merge it! Do you have permission to merge?

ceccoemi:
@lk-geimfari I'm not able to merge. I don't know if it is due to the failing tests or to the permissions.

lk-geimfari:
@ceccoemi Merged!

PR #778, opened 2019-12-07
Matrix

### LINKED ISSUE #727: The matrix generators for provider Numbers()

ceccoemi:
# I have made things!

## Checklist

## Related issues

Format is:
- Closes #727 
- Refs #727

I added the function `matrix()` to the `Numbers` provider. 

You can choose the type of the generated numbers with the argument `num_type`. There is the argument `**kwargs` that is used to pass parameters specific to the type of the generated numbers.

I don't know if I did the type annotation correctly, please let me know.

### DIFF
--- a/mimesis/enums.py
+++ b/mimesis/enums.py
@@ -218,7 +218,7 @@ class UnitName(Enum):
     RADIOACTIVITY = ('becquerel', 'Bq')
 
 
-class NumType(Enum):
+class NumTypes(Enum):
     """Provide number types.
 
     An argument for :meth:`~mimesis.Numbers.matrix()`.
--- a/mimesis/enums.py
+++ b/mimesis/enums.py
@@ -216,3 +216,14 @@ class UnitName(Enum):
     INDUCTANCE = ('henry', 'H')
     TEMPERATURE = ('Celsius', '°C')
     RADIOACTIVITY = ('becquerel', 'Bq')
+
+
+class NumType(Enum):
+    """Provide number types.
+
+    An argument for :meth:`~mimesis.Numbers.matrix()`.
+    """
+
+    FLOATS = 'floats'
+    INTEGERS = 'integers'
+    COMPLEXES = 'complexes'
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -4,7 +4,7 @@
 
 from typing import Callable, List, Union
 
-from mimesis.enums import NumType
+from mimesis.enums import NumTypes
 from mimesis.providers.base import BaseProvider
 
 __all__ = ['Numbers']
@@ -68,19 +68,19 @@ class Numbers(BaseProvider):
             for _ in range(n)]
 
     def matrix(self, m: int = 10, n: int = 10,
-               num_type: NumType = NumType.FLOATS, **kwargs) -> List[List]:
+               num_type: NumTypes = NumTypes.FLOATS, **kwargs) -> List[List]:
         """Generate a m x n matrix with random numbers.
 
         :param m: Number of rows.
         :param n: Number of columns.
-        :param num_type: NumType enum object.
+        :param num_type: NumTypes enum object.
         :param **kwargs: Other specific arguments.
         :return: A matrix of random numbers.
         """
-        key = self._validate_enum(num_type, NumType)
+        key = self._validate_enum(num_type, NumTypes)
         kwargs.update({'n': n})
-        gen_fun = getattr(self, key)
-        return [gen_fun(**kwargs) for _ in range(m)]
+        method = getattr(self, key)
+        return [method(**kwargs) for _ in range(m)]
 
     @staticmethod
     def primes(start: int = 1, end: int = 999) -> List[int]:
--- a/mimesis/providers/numbers.py
+++ b/mimesis/providers/numbers.py
@@ -4,6 +4,7 @@
 
 from typing import Callable, List, Union
 
+from mimesis.enums import NumType
 from mimesis.providers.base import BaseProvider
 
 __all__ = ['Numbers']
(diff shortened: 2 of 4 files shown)


mimesis/providers/numbers.py
```diff
@@ -66,6 +66,32 @@ def complexes(self, start_real: float = 0, end_real: float = 1,
                     self.random.uniform(start_imag, end_imag, rounding_imag))
             for _ in range(n)]
 
+    def matrix(self, m: int = 10, n: int = 10, num_type: str = 'floats',
```
lk-geimfari: It's better to create enum object:

```
def matrix(self, m: int = 10, n: int = 10, num_type: str = NumType.FLOAT)
```

I think that the idea is pretty clear.

mimesis/providers/numbers.py
```diff
@@ -66,6 +66,32 @@ def complexes(self, start_real: float = 0, end_real: float = 1,
...
+
+        num_type could be "float", "integers" or "complexes".
+
+        :param m: Number of rows.
+        :param n: Number of columns.
+        :param num_type: Type of numbers to generate.
+        :param **kwargs: Other specific arguments.
+        :return: A matrix of random numbers.
+        """
+        gen_func: Callable
```
lk-geimfari: Instead of this, use this:

```
self._validate_enum(num_type, NumType)
```

mimesis/providers/numbers.py
```diff
@@ -66,6 +66,32 @@ def complexes(self, start_real: float = 0, end_real: float = 1,
...
+        if num_type == 'floats':
+            gen_func = self.floats
+        elif num_type == 'integers':
+            gen_func = self.integers
+        elif num_type == 'complexes':
+            gen_func = self.complexes
+        else:
+            raise ValueError(
```
lk-geimfari: No need in this when you use `self. _validate_enum()`

ceccoemi: @lk-geimfari do you mean something like that?

```
def matrix(num_type: NumType = None):
    gen_func = {
        NumType.FLOATS: self.floats,
        NumType.INTEGERS: self.integers,
        NumType.COMPLEXES: self.complexes
    }
    key = self._validate_enum(item=num_type, enum=NumType)
    return gen_func[key](kwargs)
```

lk-geimfari: Not really. [There](https://github.com/lk-geimfari/mimesis/blob/master/mimesis/providers/person.py#L83) is an example.

lk-geimfari: Enums must be defined in `mimesis.enums`.

lk-geimfari: Something like this:

```
_type = self._validate_enum(num_type, NumType)

if not hasattr(self, _type):
    raise AttributeError('Number has not such provider')

return getattr(self, _type)(**kwargs)
```

ceccoemi: @lk-geimfari nice! I forgot about the `getattr` function!

ceccoemi: @lk-geimfari please check out my last commit.

I think that 

    if not hasattr(self, nm_type):
        raise AttributeError('Number has not such method')

is not needed because the argument is already checked in `_validate_enum`

mimesis/enums.py
```diff
@@ -216,3 +216,14 @@ class UnitName(Enum):
     INDUCTANCE = ('henry', 'H')
     TEMPERATURE = ('Celsius', '°C')
     RADIOACTIVITY = ('becquerel', 'Bq')
+
+
+class NumType(Enum):
```
lk-geimfari: I think that `NumTypes` sounds much better when enum contains items named in the plural.

`NumType.FLOATS` vs `NumTypes.FLOATS`. 

What do you think?

mimesis/providers/numbers.py
```diff
@@ -66,6 +67,21 @@ def complexes(self, start_real: float = 0, end_real: float = 1,
...
+
+        :param m: Number of rows.
+        :param n: Number of columns.
+        :param num_type: NumType enum object.
+        :param **kwargs: Other specific arguments.
+        :return: A matrix of random numbers.
+        """
+        key = self._validate_enum(num_type, NumType)
+        kwargs.update({'n': n})
+        gen_fun = getattr(self, key)
```
lk-geimfari: It's better to name it as `method` instead of `gen_func`. Don't you think?

lk-geimfari:
@ceccoemi Wait for Travis CI and let me know if you can merge it. I have approved this PR.

PR #779, opened 2019-12-12
FutureDatetime and PastDatetime providers

### LINKED ISSUE #775: Add future and past providers to datetime
I would love to see the same feature from `Datetime` ported to both `FutureDatetime` and `PastDatetime` providers.

I hope that you will get the idea from the class names 🙂

ceccoemi:
# I have made things!

## Checklist

- [ ] I'm sure that I did not unrelated changes in this pull request

## Related issues

Format is:
- Closes #775 
- Refs #775 

@lk-geimfari 
@sobolevn 

Before I continue any further, I'd like to know what do you think about this approach.

`FutureDatetime` is a subclass of `Datetime`. It takes some argument in the `__init__()` function to specify how much time to shift from the time when it's initialized.

All the methods that take as argument a `start` date are overrided and `start` is replaced with the current future. The `end` argument is set in a way that the timedelta from `start` to `end` is the same in `Datetime` and in `FutureDatetime`. For example in `Datetime.week_date()`, `start=2017` and `end=2018`, there is a year of difference in the default arguments, so in `FutureDatetime`, by default, we have `start=self.future.year` and `end=self.future.year+1`.


lk-geimfari:
@ceccoemi I see you have requested a review, but as I can judge there is only one method implemented in the provider. Did you request a review to illustrate the idea?

ceccoemi:
> @ceccoemi I see you have requested a review, but as I can judge there is only one method implemented in the provider. Did you request a review to illustrate the idea?

@lk-geimfari yes exactly, I don't want to write a big amount of code and then discover that you don't like my solutions. The other methods will be implemented with the same approach.

lk-geimfari:
@ceccoemi I agree. It's better to discuss solutions. Well, it looks good for me. @sobolevn What do you think? Any comments?

mimesis/providers/date.py
```diff
@@ -255,3 +254,128 @@ def timestamp(self, posix: bool = True, **kwargs) -> Union[str, int]:
...
+        self.day_of_week = self._dt.day_of_week
+        self.month = self._dt.month
+        self.century = self._dt.century
+        self.periodicity = self._dt.periodicity
+        self.time = self._dt.time
+        self.formatted_time = self._dt.formatted_time
+        self.day_of_month = self._dt.day_of_month
+        self.timezone = self._dt.timezone
+        self.gmt_offset = self._dt.gmt_offset
+        self.timestamp = self._dt.timestamp
```
ceccoemi: @lk-geimfari @sobolevn I had to switch to composition instead of inheritance because of the incompatibility with the overloaded methods.

I know this section is a bit tricky. It can be replaced with:

```
        self._dt = Datetime(*args, **kwargs)
        for method in (
            m for m in dir(Datetime)
            if callable(getattr(Datetime, m))
            and not m.startswith('_')
            and m not in dir(FutureDatetime)
        ):
            setattr(self, method, getattr(self._dt, method))
```
which it automatically adds the methods not overridden, but it is more black magic and less explicit. What do you prefer?

lk-geimfari: @ceccoemi I think that we need to implement only some methods in this provider, not all of them. We must decide which methods can be really useful in `FutureDatetime`.

ceccoemi: @lk-geimfari ok, if you don't want all the methods then it's really easy.

I propose to keep `date`, `formatted_date`, `datetime`, `formatted_datetime` and `timestamp` only.

ceccoemi: @lk-geimfari see my last commit!

lk-geimfari: @ceccoemi I agree with your propose.

ceccoemi:
@lk-geimfari please wait before publish a new version. There are some sections in the documentation that must be updated. I will do it as soon as this PR is merged in another branch.

lk-geimfari:
@ceccoemi Although I completely like your solution, I'm not quite sure that this idea is good when now I see how much code of the same type we got because of the specifics of the issue.

If you do not mind I'll try to figure out how we can do it in a much simple and clear way. Can you, please, freeze your activity on this PR for now? You can work on another issue if you want.

lk-geimfari:
@ceccoemi Sorry for this. I hope that you're not mad.

ceccoemi:
@lk-geimfari ok, no problem.

ceccoemi:
@lk-geimfari I agree that there is a lot of repetition.

Maybe the best thing to do is to implement just some methods in the `Datetime` provider. I'm thinking of `future_date()`, `future_datetime()` ...

lk-geimfari:
@ceccoemi We need to check out how the faker implements similar features and implement this feature in a similar way.

Related: https://faker.readthedocs.io/en/master/providers/faker.providers.date_time.html

PR #780, opened 2019-12-13
Add support of IP objects for the Internet provider (Fix #761)

### LINKED ISSUE #761: Improve IPv4/IPv6 generators for Internet()
We can improve how `.ip_v4()` and `.ip_v6()` works.

I suggest using `ipaddress` module from the standard library, just like this:

```python
    def ip_v4(self, with_port: bool = False, stringify: bool = True) -> Union[str, IPv4Address]:
        """Generate a random IPv4 address.

        :param stringify:  Return IPv4 as string (True by default).
        :param with_port: Add port to IP.
        :return: Random IPv4 address.

        :Example:
            19.121.223.58
        """
        ipv4 = IPv4Address(self.random.getrandbits(32))

        if not stringify:
            return ipv4

        if with_port:
            ipv4 += ':{}'.format(self.port())

        return ipv4
```

The idea is to support IP addresses as objects to make them more useful than just strings.

lk-geimfari:
I have added a new method to ensure the support of IP objects using `ipaddress` model from the standard library.

Usage:

```python
from mimesis import Internet

>>> net.ip_v4() # Also this method now support new argument — port_range
'158.184.192.44'

>>> net.ip_v4_object() # <-- This is a new feature
IPv4Address('113.69.6.167')

>>> net.ip_v6()
'8615:7b8d:4744:c9cf:18e4:16c:9bf2:7689'

>>> net.ip_v6_object() # <-- This is a new feature
IPv6Address('58f8:1d67:a62c:3b48:f1a5:16a6:14ea:2901')
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,4 +1,4 @@
-Version 3.3.1
+Version 4.0.0
 -------------
 
 .. note:: This version is still under development.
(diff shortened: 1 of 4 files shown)


lk-geimfari:
@ceccoemi Any comments?

PR #822, opened 2020-03-21
Create misspell.yml

sobolevn:
Currently it is set to `push` because I want to have the initial run right now.
Later I will change it to work once a day.


sobolevn:
Related: https://github.com/lk-geimfari/mimesis/pull/823
But, I have to ignore `.json` files.

sobolevn:
Actually, there are not typos here. So, we can ignore this check.

PR #867, opened 2020-06-05
Fix incorrect native language name for Ukrainian

webknjaz:


lk-geimfari:
@webknjaz Thank you! It must be an annoying one.

PR #876, opened 2020-06-08
Fix unicode support in Random.custom_code

eumiro:
# I have made things!

## Checklist

`random.Random.custom_code` raised `IndexError: bytearray index out of range` if `mask` contained characters beyond ASCII because the created `code` `bytearray` was shorter than the encoded `mask`.
This fixes the issue (and adds a test) that allows the whole Unicode in `mask` (still not in `char`/`digit` placeholders). In UTF-8 this is possible since characters beyond ASCII are encoded using bytes other than 32–126.

### DIFF
--- a/mimesis/random.py
+++ b/mimesis/random.py
@@ -75,7 +75,6 @@ class Random(random_module.Random):
         """
         char_code = ord(char)
         digit_code = ord(digit)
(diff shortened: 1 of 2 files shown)


eumiro:
Sorry, had to fix a typo where I compared result with result and not result with expected.

PR #974, opened 2020-11-16
Add GitHub Actions

### LINKED ISSUE #963: Migrate to Github Actions
Well, seems like Travis CI is not really free anymore, so we need to migrate to Github Actions.

michael0liver:
# I have made things!

## Checklist

- [x] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

## Related issues

- Closes #963

I made an initial attempt at migrating the current Travis CI to GitHub Actions. @lk-geimfari you will have to create some repository secrets:

- [x] `CODECOV_TOKEN`
- [ ] `PYPI_PASSWORD`

For now I have left the codecov step commented out so it doesn't fail the pipeline, once secrets are in place we can try properly.

I also cached the virtual environment so we get quicker subsequent runs if the dependencies don't change 😄

This is what it looks like:

image


.github/workflows/ci.yml
```diff
@@ -0,0 +1,56 @@
...
+          poetry python setup.py minify
+          poetry run check-manifest
+          poetry run python setup.py check --restructuredtext --strict
+      - name: Publish
+        env:
+          PYPI_USERNAME: lk-geimfari
+          PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          poetry build
+          poetry publish --username ${{ env.PYPI_USERNAME }} --password ${{ env.PYPI_PASSWORD }}
```
lk-geimfari, line 56: I'm not sure that this is secure enough. Will it be printed?

lk-geimfari:
@michael0liver First of all — thank you very much for this PR. I've added a comment.

.github/workflows/ci.yml
michael0liver: GitHub Actions will redact any secrets from the log file so it should be okay.. BUT I looked up the appropriate environment variables for `poetry` and `codecov` and used them instead.

michael0liver: Turns out that `codecov` has an [official Action](https://github.com/marketplace/actions/codecov), I had added the appropriate step.

I am not sure the current release strategy but this integration can be further improved by automatically publishing package version to PyPI  based on a GitHub Tagged Release. Right now if will publish on a `push` event to the `master` branch so this PR should probably be changed to merge to the `dev`. This would need to be discussed.

lk-geimfari: @michael0liver Thank you very much! I'll merge this PR as soon, as I will be near to my laptop.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+      # - name: Upload coverage to Codecov
+      #   uses: codecov/codecov-action@v1
+      #   with:
+      #     token: ${{ secrets.CODECOV_TOKEN }}
+      #     file: ./coverage.xml
+      #     env_vars: OS,PYTHON
+      #     fail_ci_if_error: true
+
+  deploy:
+    if: ${{ github.ref == 'ref/head/master' }}
```
lk-geimfari, line 45: Do I understand correctly that this will trigger a new release each time when changes merged to master?

.github/workflows/ci.yml
lk-geimfari: @michael0liver Can you, please, change the release strategy to make it publish only when tags are pushed? I think that this is a better strategy for us.

lk-geimfari: Also, maybe we can use `POETRY_PYPI_TOKEN_PYPI`? To publish new packages on PyPi using API token. What do you think?

.github/workflows/ci.yml
michael0liver: Yes this workflow is triggered by the `push` event but the `deploy` job will only run if the `build` job was successful AND when merging to master.

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows

We could setup a separate workflow file that handles the publishing if required.

lk-geimfari: Isn't a better solution to use tags for publishing new releases?

.github/workflows/ci.yml
michael0liver: @lk-geimfari I split the workflow files. Now we have `publish.yml` and `test.yml`.
You still have to create some secrets `PYPI_PASSWORD` and `CODECOV_TOKEN`. You can name these whatever you want but make sure to update them in workflow file.

.github/workflows/ci.yml
michael0liver: @lk-geimfari yes I changed, check new changes.

.github/workflows/ci.yml
michael0liver: > Also, maybe we can use `POETRY_PYPI_TOKEN_PYPI`? To publish new packages on PyPi using API token. What do you think?

Sounds good, I will update the environment variables.

michael0liver: @lk-geimfari Should we pin `POETRY_VERSION` in CI ?

lk-geimfari: No, I think it shouldn't

michael0liver: @lk-geimfari Okay what is outstanding to get this merged?

.github/workflows/ci.yml
lk-geimfari: Thanks! I'll merge this PR in the morning. Thank you again!

lk-geimfari (approved):

lk-geimfari:
@michael0liver Thank you again. You have done a great job!.

lk-geimfari:
@michael0liver Sure! I'll do it after solving #976 and #975. Thanks again!

PR #1008, opened 2021-01-25
Update format for series_and_number

xapoh11:
Need a space between series and number in RU passport.

# I have made things!

## Checklist

### DIFF
--- a/mimesis/builtins/ru.py
+++ b/mimesis/builtins/ru.py
@@ -82,7 +82,7 @@ class RussiaSpecProvider(BaseSpecProvider):
         :Example:
             57 16 805199.
         """
-        return '{}{}'.format(
+        return '{} {}'.format(
             self.passport_series(),
             self.passport_number(),
         )


mimesis/builtins/ru.py
```diff
@@ -82,7 +82,7 @@ def series_and_number(self) -> str:
         :Example:
             57 16 805199.
         """
-        return '{}{}'.format(
+        return '{} {}'.format(
```
lk-geimfari, line 85: Please, update test for this method.

lk-geimfari (changes requested):

mimesis/builtins/ru.py
xapoh11: Mmm, can you explane what do you mean?
I am new for that.
I saw test at https://github.com/lk-geimfari/mimesis/blob/master/tests/test_builtins/ru/test_russia_spec.py
'''
def test_series_and_number(russia):
    result = russia.series_and_number()
    assert result is not None
'''
The test does not require updating after my edit

PR #1054, opened 2021-05-10
Rework CPR generation

### LINKED ISSUE #1050: The DenmarkSpecProvider generates invalid CPR-numbers
# Bug report
The DenmarkSpecProvider generates invalid CPR-numbers

## What's wrong
The `DenmarkSpecProvider` generates CPR numbers using this code:
```
        day = '{:02d}'.format(self.random.randint(1, 31))
        month = '{:02d}'.format(self.random.randint(1, 12))
        year = '{:02d}'.format(self.random.randint(0, 99))
        serial_number = '{:04d}'.format(self.random.randint(0, 9999))

        cpr_nr = '{}{}{}{}'.format(day, month, year, serial_number)
```
This is will yield incorrect CPR numbers, for instance the CPR number: `3102509000` which is invalid for several reasons:
1. The date part should be valid; 31th of Feburary 1950/2050 is not a valid date.
2. The `serial_number` is not just a random number, but has format to it, specifically: `{century_selector}{serial_digit1}{serial_digit2}{checksum}`.
3. The `century_selector` should be 0-4 if the person was born from 1900-1999, and 5-9 if they were born in 2000-2057.
As 2050 has not happened yet, the year for this CPR number must be 1950, and thus the `century_selector` should have been 0-4 instead of 9.
4. The `checksum` is invalid, in this case it should have been `4`, and should be calculated according to the official spec.

## How is that should be

The generated `day`, `month` and `year` should together constitute a valid date between 1900 and 2057.
The `century_selector` should be derived from the `year` according to the current year.
The `serial_number` (2 middle digits) can be completely randomly generated.
The `checksum` number should be computed according to the official spec.
- Strictly speaking the `checksum` is no longer enforced after October 2007, and thus a completely random digit is actually acceptable, but is exceedingly rare compared to previously valid `checksum` numbers.

Additionally it would make sense that the method took a `Gender` enum value as an argument, as the parity of the last digit can be used to determine gender, with even being females.

Skeen:
# Rework CPR generation

This PR reworks CPR generation within the DenmarkSpecProvider, as discussed on #1050.

## Checklist

- [X] I have read [contributing guidelines](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTING.rst)
- [X] I'm sure that I did not unrelated changes in this pull request
- [X] I have created at least one test case for the changes I have made
- [X] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)
- [X] I have added my changes to the [CHANGELOG.rst](https://github.com/lk-geimfari/mimesis/blob/master/CHANGELOG.rst)

## Related issues

Closes #1050.

### DIFF
--- a/mimesis/builtins/da.py
+++ b/mimesis/builtins/da.py
@@ -1,9 +1,11 @@
 # -*- coding: utf-8 -*-
 
 """Specific data provider for Denmark (da)."""
-
+import operator
+from typing import Tuple
 from typing import Optional
 
+from mimesis import Datetime
 from mimesis.builtins.base import BaseSpecProvider
 from mimesis.locales import Locale
 from mimesis.typing import Seed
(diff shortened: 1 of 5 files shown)


lk-geimfari:
@Skeen Currently, all the work done in `dev`, because I'm preparing the 5.0.0 version. Can you please resolve conflicts caused by the change of a base?

mimesis/builtins/da.py
```diff
@@ -14,26 +16,83 @@ class DenmarkSpecProvider(BaseSpecProvider):
...
+            return self.random.randint(5, 8)
+        elif 1900 <= year < 1937:
+            return self.random.randint(0, 3)
+        elif 1937 <= year < 2000:
+            return self.random.choice([4, 9])
+        elif 2000 <= year < 2037:
+            return self.random.randint(4, 9)
+        # elif 2037 <= year < 2058:
+        #     return self.random.randint(5, 9)
+        raise ValueError('Invalid year')
```
lk-geimfari, line 39: This line is not covered by the test.

lk-geimfari:
@Skeen ping

mimesis/builtins/da.py
Skeen: Fixed

Skeen:
> @Skeen Currently, all the work done in `dev`, because I'm preparing the 5.0.0 version. Can you please resolve conflicts caused by the change of a base?

Rebased

PR #1069, opened 2021-07-26
Fix publishing pipeline

hoefling:
Reason: since v4.1.1, no wheels are being released anymore. The change in #820 is responsible for that. If you look at the [publishing pipelines](https://github.com/lk-geimfari/mimesis/actions/workflows/publish.yml), those fail b/c `poetry python X` is not a valid command. This PR amends that.

## Checklist

- ~~[ ] I have created at least one test case for the changes I have made~~ only changes in CI config

## Related issues

None

@lk-geimfari right now, I see the tag `5.0.0` in the repo, but there were no dists pushed to PyPI on merging #1035. I'm not sure if this was a deliberate decision or the pipeline just failed and no one looked at it. If the latter - is it possible to release a `5.0.0.post1` or `5.0.1` which has the same codebase as `5.0.0`, only with wheels on PyPI?


lk-geimfari:
@hoefling Although we have created the 5.0.0 release, we are not publishing it on PyPi yet, since were found some critical issues.

lk-geimfari (approved):

lk-geimfari:
Thank you!

hoefling:
@lk-geimfari thank you for your response! Don't merge this in yet though, there's still stuff to do (like `MANIFEST.in` obsoletion etc). Will do later this evening and ping you when done.

lk-geimfari:
@hoefling I'll wait for ping. Thank you!

pyproject.toml
```diff
@@ -4,6 +4,43 @@ version = "5.0.0"
 description = "Mimesis: fake data generator."
 authors = ["Isaak Uchakaev <likid.geimfari@gmail.com>"]
 license = "MIT"
+readme = "README.rst"
+homepage = "https://github.com/lk-geimfari/mimesis"
+repository = "https://github.com/lk-geimfari/mimesis"
+documentation = "https://mimesis.readthedocs.io"
```
lk-geimfari, line 10: Please, change it to https://mimesis.name

lk-geimfari (approved):

pyproject.toml
hoefling: @lk-geimfari done!

hoefling:
@lk-geimfari I would also rewrite the minifier to 1. get rid of `setup.py` and 2. get rid of using `distutils` API since deprecated and the minifier doesn't really need any of that. However, this could be food for a separate PR, as the changes get too big otherwise.

hoefling:
@lk-geimfari other than that, I think this is ready for a review now!

lk-geimfari:
@hoefling Okay then. I'm merging this PR. Thank you very much!

PR #1129, opened 2021-11-26
Fix types for Seed.

ngnpope:
Makes the following changes:

- Added `float` to the `Union` for `Seed`.
- Added `None` to the `Union` for `Seed`.
- Replaced `Optional[Seed]` with `Seed` everywhere.

See the following links for details:

- https://bugs.python.org/issue32554
- https://github.com/python/cpython/pull/15382
- https://docs.python.org/3.9/library/random.html#random.seed

Currently `typeshed` is still using `Any`.

## Checklist

- [n/a] I have created at least one test case for the changes I have made

### DIFF
--- a/mimesis/builtins/da.py
+++ b/mimesis/builtins/da.py
@@ -2,7 +2,7 @@
 
 """Specific data provider for Denmark (da)."""
 import operator
-from typing import Final, Optional, Tuple
+from typing import Final, Tuple
 
 from mimesis import Datetime
 from mimesis.builtins.base import BaseSpecProvider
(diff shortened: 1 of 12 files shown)


lk-geimfari:
@ngnpopePlease, fix the issues (see details for github actions)

ngnpope:
Sure. (I'd noticed but just don't have time to get back to it this weekend. Will resolve in a few days.)

ngnpope:
Hopefully I've fixed all the issues. Sorry for the laziness and the noise!

lk-geimfari:
@ngnpope You did a great job! Thank you very much!

Issue #1150, opened 2022-01-06
Generic provider does not have a per instance random generator.


ngnpope:
# Bug report

## What's wrong

- `BaseProvider` calls `.reseed()` from `.__init__()` but only when `seed is not None`.

  https://github.com/lk-geimfari/mimesis/blob/01a29e73eb4f4d9c80bea22f782d42754f8647e1/mimesis/providers/base.py#L24-L53

- `Generic` overrides `.reseed()` and reseeds its sub-providers. However, it doesn't call `super().reseed(seed)` which means that it never creates its own per-instance `random`, nor does it reseed it. This means that the random instance on `Generic` cannot be used reproducibly.

  https://github.com/lk-geimfari/mimesis/blob/01a29e73eb4f4d9c80bea22f782d42754f8647e1/mimesis/providers/generic.py#L113-L125

## How is that should be

- `BaseProvider.reseed()` should always be called from `BaseProvider.__init__()`, even when `seed` is `None` to ensure that a per-instance `random` is added. _(I don't see why you wouldn't always want this.)_ And would it then make sense to always do `self.random = Random()` in `BaseProvider.__init__()` and not in `BaseProvider.reseed()`?

  ```python
  class BaseProvider:
      ...
  
      def __init__(self, *, seed: Seed = None, **kwargs: Any) -> None:
          self.seed = seed
          self.random = Random()
          self.reseed(seed)
  
      def reseed(self, seed: Seed = None) -> None:
          self.seed = seed
          self.random.seed(seed)

      ...
  ```
- `Generic.reseed()` should call `super().reseed(seed)`.

lk-geimfari:
@ngnpope Can you please create PR? I would happy to merge it and create a new release.

PR #1155, opened 2022-01-07
Fixed providers to use isolated `random` instance with seed of `None`.

ngnpope:
Originally a dedicated random instance was used for `BaseProvider` when it was changed in c44bd931a05b871783771af7152f1217dd8b8185. It should be noted that the random instance was not seeded when passing `None`, but this was in effect irrelevant because the default seed it `None` and the following:

```python
        self.random = Random()
        if seed is not None:
            self.random.seed(seed)
```

Could have been written as the following and worked the same:

```python
        self.random = Random(seed)
```

Then c9195d72e231a98cf42aa9d5226cd136ce62ba2d changed how this worked where the seed was `None`. In `__init__()` the value of `self.random` was set to a global `mimesis.random.random` instance and a `.reseed()` method was added that is only called from `__init__()` when the seed is not `None`. Subsequent manual calls of `.reseed(None)`, however, will change `self.random` to a new instance if it is found to be that global instance.

`None` is a valid seed, and it isn't apparent why it there is this special behaviour unless `None` is being treated as "not set" in which case a separate `unset` sentinel object may be warranted.

When this change was made there was a related link to #469 which was something to do with `pytest-randomly`. I'm not sure whether the intent was to have a default state whereby the seed could be changed for the global instance which would propagate down to all providers in tests as long as the seed is `None`. If this is the case then this change won't work.

# I have made things!

## Checklist

## Related issues

Refs #1150.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ Version 5.3.0
 
 - Fix reseeding of the random generator of ``Generic``. This was a regression in v5.1.0. (See `#1150 <https://github.com/lk-geimfari/mimesis/issues/1150>`_).
 - ``Development.version()`` now supports use of both the ``calver`` and ``pre_release`` flags together.
+- Providers now have an isolated ``random`` instance when using a seed of ``None``.
 
 
 Version 5.2.1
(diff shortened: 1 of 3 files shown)


lk-geimfari:
I'm not sure about this one. I have to check this PR closely. Although you're right (it makes no sense if `seed` is `None` anyway).

lk-geimfari:
Well, seems like a good idea, so I've merged this PR.

PR #1176, opened 2022-01-20
Update `[build-system]`

sobolevn:
I try to run https://github.com/hauntsaninja/mypy_primer on this project, but the build system is quite old.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,6 @@ default_section = "FIRSTPARTY"
 [tool.taskipy.tasks]
 minify = "python minifier.py"
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0"]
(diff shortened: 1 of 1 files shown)


sobolevn:
```
 ______________________ TestInternet.test_query_string[8] _______________________

self = <tests.test_providers.test_internet.TestInternet object at 0x7f6d503de0a0>
net = <mimesis.providers.internet.Internet object at 0x7f6d503ded00>, length = 8

    @pytest.mark.parametrize("length", [4, 6, 8, None])
    def test_query_string(self, net, length):
        query_string = net.query_string(length)
        query_params_count = len(query_string.split("&"))
    
        if length is None:
            assert 1 <= query_params_count <= 10
        else:
>           assert query_params_count == length
E           assert 7 == 8

tests/test_providers/test_internet.py:118: AssertionError
```

cc @lk-geimfari

lk-geimfari:
I think that it happens because of duplication of words in `self.query_string`. I'm about to fix that behavior as soon as possible.

PR #1229, opened 2022-06-22
Avoid infinite loop in `Datetime.bulk_create_datetimes` by checking for positive timedelta

Wooza:
Hi,

we were evaluating our [automated unit test generator](https://github.com/se2p/pynguin) on mimesis and found a bug in `Datetime.bulk_create_datetimes`. It is possible to call this method without `**kwargs` or with `**kwargs` that represent a non-positive timedelta. In such a case, the `while` loop in this method never terminates. I have added a simple check to verify that the used timedelta is positive and also added a test case for it.

## Checklist

- [x] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

Best regards
Florian

### DIFF
--- a/tests/test_providers/test_datetime.py
+++ b/tests/test_providers/test_datetime.py
@@ -45,6 +45,11 @@ class TestDatetime:
         with pytest.raises(ValueError):
             _datetime.bulk_create_datetimes(None, None)
 
+        with pytest.raises(ValueError):
+            _datetime.bulk_create_datetimes(
+                date_start, date_start + datetime.timedelta(days=1)
+            )
+
     def test_year(self, _datetime):
         result = _datetime.year(minimum=2000, maximum=_datetime.CURRENT_YEAR)
         assert result >= 2000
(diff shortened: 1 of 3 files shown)


tests/test_providers/test_datetime.py
```diff
@@ -45,6 +45,11 @@ def test_bulk_create_datetimes_error(self, _datetime):
         with pytest.raises(ValueError):
             _datetime.bulk_create_datetimes(None, None)
 
+        with pytest.raises(ValueError):
```
lk-geimfari, line 48: I don't get it. Why error should raise here if timedelta is positive in this particular case?

lk-geimfari: Oh, I get it. It just incrementing `date_start` by `timedelta(0)` that make no sense.

lk-geimfari (approved):

PR #1114, opened 2021-11-02
Add "allow_random" in country() method signature

spalmieri-shippo:
## Checklist

- [x] I have read [contributing guidelines](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTING.rst)
- [X] I'm sure that I did not unrelated changes in this pull request
- [X] I have created at least one test case for the changes I have made

- [X] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

- [X] I have updated the documentation for the changes I have made

## Related issues

 Closes #1113

PR #1251, opened 2022-08-14
Add allow_random param in country method of Address provider

### LINKED ISSUE #1248: Address provider function country() does not apply locale or support allow_random param
# Bug report

## What's wrong

The [API documentation](https://mimesis.name/en/master/api.html#mimesis.Address.country) for `Address.country()` indicates that it will "Get the country of the current locale", but the current local is not applied, and it instead always returns a random country:

```
>>> en = mimesis.Address(locale=Locale.EN)
>>> en.country()
'Nicaragua'
>>> en.country()
'Venezuela'
>>> en.country()
'Fiji'
```

Additionally, the function docs include the parameter `allow_random`. However, this has not been implemented:

```python

>>> en.country(allow_random=False)
Traceback (most recent call last):
  File "/Users/willthompson/.pyenv/versions/3.9.11/lib/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
TypeError: country() got an unexpected keyword argument 'allow_random'
```

## How is that should be

Behavior should align with the documentation:
- Locale should always be applied by default
- Function should accept an `allow_random` parameter that ignores the locale when set to `False`

It looks like this was already implemented in #1114, but closed before merging. Could that be reopened and fixed up?

willsthompson:
# I have made things!

## Checklist

## Related issues

- Closes #1248

I copied the relevant changes from #1114 and credited the original author in the commit. Please let me know if you want any other changes.

### DIFF
--- a/mimesis/providers/address.py
+++ b/mimesis/providers/address.py
@@ -183,14 +183,18 @@ class Address(BaseDataProvider):
         key = self.validate_enum(code, CountryCode)
         return self.random.choice(COUNTRY_CODES[key])
 
-    def country(self) -> str:
+    def country(self, allow_random: bool = False) -> str:
         """Get the country of the current locale.
 
-        :allow_random: Return a random country name.
+        :param allow_random: Return a random country name.
         :return: The Country.
         """
-        countries: t.List[str] = self.extract(["country", "name"])
-        return self.random.choice(countries)
+        if allow_random:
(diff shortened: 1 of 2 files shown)


spectrumvoip-stefano:
@willsthompson Thanks for crediting me! Took me a while to come back to this 😄

Issue #1320, opened 2023-03-02
NSFW words generated by `Text.words` in `en`


mhauru:
The English `text.json` file (https://github.com/lk-geimfari/mimesis/blob/master/mimesis/data/en/text.json) has some words in the `"normal"` section that should probably be in the `"bad"` section. Some candidates for moving to `"bad"` that I spotted while browsing the list: "cock", "cocks", "pussy", "blowjob", "blowjobs", "booty", "busty", "butt". I only quickly browsed the file, that list is probably very incomplete.

There are also words that are in both `"bad"` and `"normal"`, although I'm not quite sure why some of them are in `"bad"`, e.g. "camel".

I haven't checked any of the other localisations, other than `en`.

lk-geimfari:
This is pretty bad. Can you please create PR with fixes?

jordiortegon:
Can confirm that there are some `"bad"` words under the `"normal"` category in Spanish too.

PR #1322, opened 2023-03-06
Move some NSFW language to "bad" words in locale en

mhauru:
Partially addresses, but does not close, #1320

I only moved the ones I easily found or previously ran into. A proper audit of the whole list might still be in order, I bet there are still some in the wrong category.


lk-geimfari:
Thank you!

PR #1324, opened 2023-03-30
Simplify `isort` configuration

sobolevn:
It is the same thing :)


sobolevn:
Oh, it actually fixed a bug that `mimesis` imports were mixed with other ones.

lk-geimfari (approved):

Issue #1313, opened 2023-01-31
Reseeding doesn't return consistent data for some providers


Svenito:
# Bug report

## What's wrong

`generic.file.file_name()` does not provide the same data when reseeding the provider.

```
In [5]: g = Generic()

In [8]: g.reseed(1)

In [9]: g.file.file_name()
Out[9]: 'conducted.md'

In [10]: g.reseed(1)

In [11]: g.file.file_name()
Out[11]: 'encouraged.md'
```

## How is that should be

When reseeding the values returned should be the same. 
It seems that the extension remains the same on each reseed but the filename differs each time.
Not all providers have this issue. So far I've only noticed this happening with `file.file_name`

## System information

mimesis 7.0.0 
python 3.9.9 
windows 11

lk-geimfari:
This is probably a bug. I'll to fix this. Thank you for report!

lk-geimfari:
This problem requires quite a serious overhaul of some providers.

This is because some providers under the hood use other providers (for example [Internet](https://github.com/lk-geimfari/mimesis/blob/master/mimesis/providers/internet.py#L49)) and when the ".reseed()" method is called it reseeding only the original provider.

lk-geimfari:
@sobolevn I apologize for the inconvenience, but do you have any eligant solution for this?

Svenito:
I hope you don't mind me making a couple of suggestions here:

One option would be to override the `reseed` method for any class that has sub-providers, and reseed those providers there.

Another option would be to iterate over the attributes of a class in the reseed base class method, and reseed any attribute that is a subclass or instance of `BaseProvider`

The latter might not be as clean, but it would help prevent human error of forgetting to reseed a sub-provider. Plus only requires a change to the `BaseProvider`

Just some ideas to throw into the ring.

lk-geimfari:
I fixed the issue only for **File**, but the bug itself is still there, so I'll reopen this issue.

Svenito:
Hi, I tentatively submitted a PR that reseeds all related providers on the BaseProvider.
Hopefully this is an acceptable way to solve the issue

lk-geimfari:
I've fixed this bug. Thank you, @Svenito.

PR #1336, opened 2023-04-01
Reseed sub providers when reseeding a provider

Svenito:
This patch will iterate over the subproviders on a provider and reseed them. Otherwise providers that rely on other providers to generate data will not return consistent data when reseeded with the same value.

Refs #1313

# I have made things!

## Checklist

- [ x ] I have read [contributing guidelines](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTING.rst)
- [ x ] I'm sure that I did not unrelated changes in this pull request
- [ x ] I have created at least one test case for the changes I have made

- [ x ] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

### DIFF
--- a/mimesis/providers/base.py
+++ b/mimesis/providers/base.py
@@ -44,6 +44,14 @@ class BaseProvider:
         :param seed: Seed for random.
             When set to `None` the current system time is used.
         """
+        providers = [
+            sub_provider
+            for _, sub_provider in self.__dict__.items()
+            if isinstance(sub_provider, BaseProvider)
+        ]
+        for p in providers:
+            p.reseed(seed)
+
         self.seed = seed
         if seed is MissingSeed:
             # Remove casts after mypy will fix this inference:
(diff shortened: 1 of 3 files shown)


mimesis/providers/base.py
```diff
@@ -44,6 +44,14 @@ def reseed(self, seed: Seed = MissingSeed) -> None:
         :param seed: Seed for random.
             When set to `None` the current system time is used.
         """
+        providers = [
+            sub_provider
+            for _, sub_provider in self.__dict__.items()
+            if isinstance(sub_provider, BaseProvider)
+        ]
+        for p in providers:
```
lk-geimfari, line 52: The problem here is thath `p` may also have attributes which are instances of `BaseProvider`.

Svenito: Wouldn't this call recurse down and call this reseed for each `p` (unless `p` overrides `reseed` and doesn't call its super).

lk-geimfari: Please see #1337

lk-geimfari:
I'm not merging this one because there is much straightforward way to do this: #1337.

No offense, please.

Svenito:
Your solution is much cleaner. Agreed. Thank you for fixing the issue.

PR #1366, opened 2023-05-10
[~] Tests: Added equal comparison to assert on test_random/test_weigh…

jordiortegon:
Added an equal operator since we could get 20 and 80 as a value from the randomize, which causes the test to fail.

## Checklist


jordiortegon:
@lk-geimfari this one should be merged before https://github.com/lk-geimfari/mimesis/pull/1365

PR #1475, opened 2024-01-26
Moved pytest-mimesis to mimesis.plugins.pytest

lk-geimfari:


lk-geimfari:
@sobolevn Speaking about [pytest-mimesis#81](https://github.com/pytest-dev/pytest-mimesis/issues/81), did I get right that now we just have to register module `mimesis.plugins.pytest` as entrypoint in pyptoject of pytest-mimesis.

Sorry for dummy question, I'm not good at this packaging related stuff.

sobolevn (commented):
Otherwise, it is impossible to review
Снимок экрана 2024-01-26 в 19 42 49

lk-geimfari:
@sobolevn Done. I've added it to ignore.

PR #1476, opened 2024-01-26
Moved pytest-mimesis to mimesis.plugins.pytest

lk-geimfari:


lk-geimfari:
@sobolevn Codecov doesn't care about ignored files. You can always just [toggle them on and off](https://docs.codecov.com/docs/github-checks#disabling-github-checks-patch-annotations) with the "a" key.

sobolevn (approved):
Thanks, TIL :)

PR #1479, opened 2024-01-28
Pytest plugin tests

lk-geimfari:


lk-geimfari:
I don't know why, but for some reason test runner do not seeing plugin. It raises `fixture 'mimesis' not found`. Am I missing something, @sobolevn?

sobolevn:
Sorry for the long review :)
Glad that it worked! 👍

lk-geimfari:
Don't worry, it's all good! I'll release new version as soon, as I finish with support of `Fieldset` fixture. By default pytest-mimesis where using only `Field`.

PR #1587, opened 2024-09-08
adding a small arabic dataset to mimesis giving it a push to start supporting arabic

yah04dev:
adding a small arabic dataset to mimesis giving it a push to start supporting arabic

# I have made things!

## Checklist

- [ ] I have created at least one test case for the changes I have made

- [ ] I have added myself to the [CONTRIBUTORS.rst](https://github.com/lk-geimfari/mimesis/blob/master/CONTRIBUTORS.rst)

## Related issues

I added a small arabic dataset but i was not able to make it working "Integrate it" because I am still a begginer, and i had a remark and an issue : arabic is an ltr language and utf8 , and I wanted to make an arabic adress format (state name <- street name <- house number) the data inerted but I wasnt able to make this format also arabic has multiple dailects and cultures ;I did my best to make the data look real I wish it will.


lk-geimfari:
@yah04dev As-salamu alaykum. First of all, thank you for contributing to the Mimesis project. I believe that adding Arabic will be a great enhancement.

However, before we can merge your Pull Request, I have a few comments:

1. Since Arabic is the official language in many countries, you need to explicitly specify for which country you are adding the data.
2. You need to add a base dataset for the Arabic language that will be suitable for use across all Arabic-speaking countries.
3. Additionally, you should create a dataset that inherits from the base one and extends it with data specific to the country for which you are adding the information.

You can find the list of language codes here: https://github.com/lk-geimfari/mimesis/tree/master/mimesis/datasets/locale_template

You can see an example of how this works with `en`, `en-ca`, and `en-au`.

lk-geimfari:
@yah04dev Overall, you should review the link I sent above, because simply adding the data, unfortunately, is not enough.

yah04dev:
@lk-geimfari Fruits , texts , quotes and some other data will stay the same because they are related to language and not culture ,So I will try to give at least two local-datasets , Thank you !

PR #1718, opened 2026-07-25
fix: resolve field aliases before custom handler lookup

### LINKED ISSUE #1514: A custom handler with an alias returns an error when using the alias
# Bug report

Repro:
field.register_handler("bloop", my_field_handler)
field.aliases.update({"zoop": "bloop"})
field("zoop") # generates error

Sanjays2402:
# I have made things!

An alias pointing at a registered custom field handler raised `FieldError`: `perform()` checked `self._handlers` with the raw name, and only `_lookup_method()` resolved aliases — so an aliased custom handler never matched and fell through to the provider lookup. Alias resolution now happens in `perform()` before the handler check, so aliases work for both built-in fields and custom handlers.

## Checklist

## Related issues

- Closes #1514

### DIFF
--- a/mimesis/schema.py
+++ b/mimesis/schema.py
@@ -122,9 +122,6 @@ class BaseField:
         :return: Callable object.
         :raise FieldError: When field is invalid.
         """
-        # Check if the field is defined in aliases
-        name = self.aliases.get(name, name)
-
         # Support additional delimiters
         name = re.sub(r"[/:\s]", ".", name)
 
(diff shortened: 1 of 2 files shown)


lk-geimfari:
@Sanjays2402 Hi! Please, see failing CI pipelines.

Sanjays2402:
Sorry for the slow turnaround — pushed the fix in 21cb2f3.

The failure was my own doing, not the aliasing change: `test_field_aliasing_with_custom_handler` registered `bloop` on the module-scoped `default_field` fixture and set `aliases`, then never cleaned up. With pytest-randomly that only bites when it happens to run before `test_unregister_all_handlers`, which is why it was red on some interpreter jobs and green on others (`assert 4 == 3` in a test my PR never touched).

It now restores state in a `finally`, so the handler and aliases are removed regardless of assertion outcome.

The new run is sitting on `action_required` — needs a maintainer to approve workflows for the branch before CI can confirm.

